Issue #6, opened 2022-02-05
lua/user/settings.lua plugins not loading


aikakira:
Hello!

Thanks for this awesome setup, it's amazing how pretty you've made everything! I'm struggling with adding new plugins via the user/settings.lua file. I can't seem to make packer sync my additional plugins, if I add them to lua/core/defaults.lua everything works and installs.

P.S Fresh install here

kabinspace:
@aikakira can you show me the settings.lua where you put your plugins once? i tried to look if something was wrong but I couldn't find any, maybe i can get some clues from you.

nielsk:
I have the same problem. Mine looks like this, I just added fugitive (because I do not yet under Gitsigns and apparently it integrates with fugitive) and the color-theme Ayu:

```
local opts = { noremap = true, silent = true }
local map = vim.api.nvim_set_keymap
local set = vim.opt

local config = {

  -- Set colorscheme
  colorscheme = "onedark",

  -- Add plugins
  plugins = {
    -- { "andweeb/presence.nvim" },
    -- { "tpope/vim-fugitive" },
    -- { "ayu-theme/ayu-vim" },
    -- {
      -- "ray-x/lsp_signature.nvim",
      -- event = "BufRead",
      -- config = function()
        -- require("lsp_signature").setup()
      -- end,
    -- },
  },

  -- On/off virtual diagnostics text
  virtual_text = true,

  -- Disable default plugins
  enabled = {
    nvim_tree = true,
    lualine = true,
    lspsaga = true,
    gitsigns = true,
    colorizer = true,
    toggle_term = true,
    comment = true,
    symbols_outline = true,
    indent_blankline = true,
    dashboard = true,
    which_key = true,
    neoscroll = true,
    ts_rainbow = true,
    ts_autotag = true,
  },
}

-- Set options
set.relativenumber = true

-- Set key bindings
map("n", "<C-s>", ":w!<CR>", opts)

-- Set autocommands
vim.cmd [[
  augroup packer_conf
    autocmd!
    autocmd bufwritepost plugins.lua source <afile> | PackerSync
  augroup end
]]

return config
```

nielsk:
And maybe it is just a layer8-problem - after I added the plugins, what would I need to do?

nielsk:
I also just set neoscroll to false and it seems not to change anything

nielsk:
I deleted the -- in the plugins-part and that worked for me. Oh the "--" are comments 🤦 
[...]

PR #7, opened 2022-02-05
Remove -- in plugins-part, so that Packer can recognize them

nielsk:
This solves the problem that PackerSync won't recognize newly added packages.
See https://github.com/kabinspace/AstroVim/issues/6

### DIFF
--- a/lua/user/settings.lua
+++ b/lua/user/settings.lua
@@ -9,14 +9,14 @@ local config = {
 
   -- Add plugins
   plugins = {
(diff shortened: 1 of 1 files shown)


nielsk:
I closed it because I actually removed only comment-signs.

kabinspace:
Sorry for the late response, but the reason I commented out those lines was I actually wanted to show that as just an example. If I would have left them uncommented, those extra plugins would be downloaded during the installation process.

PR #16, opened 2022-02-06
README.md Updated with License

iqbalpb01:


kabinspace:
I appreciate your effort, but the info about the license is already given as you can see in the picture below
alreadydoneee

iqbalpb01:
Thank You

PR #18, opened 2022-02-07
Fix the icon of "New File" on the dashboard

MagicalLiebe:
Thank you for publishing a very useful and cool project.

I found something wrong about the "New File" icon on the dashboard.
As you can see in the video below, the behavior of the cursor on the "New File" is different from the other items.
The cursor moves to the second letter "e" instead of the first letter "N".

https://user-images.githubusercontent.com/38152917/152708229-7160650c-3835-4bd1-b3c1-a80532d899fc.mp4

So I replaced it with the correct "plus" icon of Nerd Font.
It now works correctly, as shown in the following video.

https://user-images.githubusercontent.com/38152917/152708234-2b3d2180-1082-414c-ae6b-56a82ee5d543.mp4


kabinspace:
Great, not a big issue but does look like something that needs to be addressed, Thank you for your contribution :)

Issue #28, opened 2022-02-08
lazygit not work


lifesign:
i have lazygit installed on my mac, but when i run <Space> + g + g, got below error

```
E5108: Error executing lua [string ":lua"]:1: attempt to call global '_LAZYGIT_TOGGLE' (a nil value)
stack traceback:
        [string ":lua"]:1: in main chunk
```

srRyo:
@lifesign I fixed this by removing this line in "core/plugins.lua"

https://github.com/kabinspace/AstroVim/blob/97509bd29e3f60a77cc2f0944fd4b76ed954e10e/lua/core/plugins.lua#L286

then run `:PackerSync` and with this it should work :)

kabinspace:
@lifesign This has been already addressed [here](https://github.com/kabinspace/AstroVim/issues/15#issuecomment-1030742915)

lifesign:
got it. thx

Issue #15, opened 2022-02-06
Terminal toggles not working


thedandroid:
None of the terminal toggles, i.e. lazygit, python, node, htop are working getting the following error:

`Error executing lua [string ":lua"]:1: attempt to call field '_HTOP_TOGGLE' (a nil value)`

kabinspace:
@thedandroid These won't work unless you first open a normal terminal. For example: You should at least open the toggleterm once like this: `Ctrl` + `\` then you can close the terminal or keep it if you like and your lazygit, python, node and htop should be working fine, The reason of this is I have lazy-loaded toggleterm for faster startup time, it won't open unless you command It to. Snce the toggleterm is not loaded at first using those python,node etc commands gives an error. You can also solve this problem if you remove [this](https://github.com/kabinspace/AstroVim/blob/0fdf5b48ebfa2008cf60697db6cc3f2032b8b136/lua/core/plugins.lua#L286) line located in `lua/core/plugins.lua` file

mehalter:
@thedandroid just a heads up I have a PR open that will resolve this issue and allow for the lazy loading to be enabled and these toglges work #241

PR #29, opened 2022-02-08
Fix: issue #28

srRyo:
### FIXED:

- [x] issue #28 

### Added:

- "For-loop" in "core/options.lua" to make it easier to add and modify options (can also be used in "user/settings.lua")

- "packer_compiled.lua" and the "/user" directory to the .gitignore file.

Note: I fully added the "user" directory, in case any user needs to create directories to keep their custom settings in order and this is not detected as a change.


kabinspace:
I really appreciate your effort but here are some of the things I want to clear, [This](https://github.com/kabinspace/AstroVim/pull/29/commits/94e69c2a11c6d61287add4cbd79be9e52e7d93bd) change is not really necessary because it's better for the user to re-compile after an update. [Here](https://github.com/kabinspace/AstroVim/pull/29/commits/e2cf96082d28798f637a5c9c7be24bc903000134) I purposely avoided using loop because it was not really necessary and I wanted to make it simple. [Here](https://github.com/kabinspace/AstroVim/pull/29/commits/b7b8c9588bb90f04b1271a719695fa83e2dfdb9f) I am only letting `.lua` because AstroVim only uses lua for the most part. [Here](https://github.com/kabinspace/AstroVim/pull/29/commits/4a7a393ae9c6c4c37a7b5e97c09be666ed05c315) toggleterm is lazyloaded for better startup time. I've already given [solution](https://github.com/kabinspace/AstroVim/issues/15#issuecomment-1030742915) in case the user's don't want to lazy load the plugin.

kabinspace:
On second thought, I think allowing users to have a directory does make sense indeed. [Here](https://github.com/kabinspace/AstroVim/pull/29/commits/b7b8c9588bb90f04b1271a719695fa83e2dfdb9f) can you please remove the blank space in line number 3 before I merge that.

srRyo:
@kabinspace Of course, I have already removed the blank space.

kabinspace:
Can you also remove the rest of the files except for [this](https://github.com/kabinspace/AstroVim/pull/29/commits/93f13b66f93d73272e5284d4c66808841d6b1355) one

kabinspace:
@srRyo Looks like you are creating conflicts with the main branch. Can you close this pull request and make another one where you only commit the changes for .gitignore?

PR #34, opened 2022-02-10
Dev

smeyerhot:
So I made these two changes since the other unicode characters don't show up on mac. Dunno if you want this change but I think it looks good.


bryant-the-coder:
Try using nerd font. Go to my repo there is a nerd font repo

kabinspace:
@smeyerhot This issue won't come as long as you use nerd fonts despite your OS. Make sure you are using nerd fonts as mentioned in the readme.

PR #35, opened 2022-02-10
Allow overriding treesitter options

0xIDANT:
This PR allows overriding the treesitter options to give the user more control.

The main purpose for this was to be able to define the parsers that should be installed in the configuration file without having to install them via command. This allows a faster and easier setup of Neovim on new/other machines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,13 @@ plugins = {
   },
 },
 
+-- Overrides
+overrides = {
+  treesitter = {
+    ensure_installed = {"go"}
+  }
+}
+
 -- On/off virtual diagnostics text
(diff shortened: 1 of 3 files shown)


kabinspace:
I appreciate your effort, I want the markdown to be untouched cause  I don't want to make it any longer. Can you remove the markdown changes

kabinspace:
On second thought, i think it's good to include it in markdown, thank you for your contribution :)

Tilley:
@kabinspace I'm keen to make more of the configs customisable. Is this the preferred pattern, and would that work be useful for this project?

PR #38, opened 2022-02-10
Packer move

hunger:
My config directory is read-only, so Packer fails in AstroVim as it can not write its `packer_compiled.lua` file.

So here is a PR to make that file location configurable. The default is of course the location you currently use:-)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -22,11 +22,10 @@ end
 
 local config = utils.user_settings()
 
-
 if type(config.polish) == "function" then
   config.polish()
 else
-  error("The polish value in your user configuration must be a function")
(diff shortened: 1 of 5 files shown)


hunger:
The first patch basically reads all the user settings once when core.utils is required for the first time, combines the data and then stores the resulting table. That *might* actually speed up access to user settings, especially with HDD... but it will probably not do much as lua will probably cache required modules anyway.

*But* I need to move the code somewhere out of the table, so that I can access those settings in the next patch:-)

hunger:
Latest version is pre-commit clean.

Issue #39, opened 2022-02-10
Can I disable the tab bar somehow?


hunger:
I hate tabs in editors with a vengeance, so can I disable that somehow?

The relevant plugin seems to be `bufferline`, but I can not find a flag for that in the settings.lua.

PR #40, opened 2022-02-10
Add option to disable bufferline

hunger:
Closes: #39

I think it is safe to remove the `after = "bufferline"` from `lualine`: I could not spot any mention of bufferline in the lualine config.

The history of plugins.lua unfortunately gave no hint why this ordering was considered encessary.

### DIFF
--- a/lua/core/defaults.lua
+++ b/lua/core/defaults.lua
@@ -7,6 +7,7 @@ local config = {
   virtual_text = true,
 
   enabled = {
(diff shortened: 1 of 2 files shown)


kabinspace:
I appreciate your effort, I never really thought about adding a disable functionality for bufferline since it is used by the majority, but it's not a bad idea to make it an option. Thank you for your contribution :)

PR #46, opened 2022-02-11
Disable Ex mode

hunger:
Ex-mode is not widely used and it is hard to get out off again,
so disable it by default.


hunger:
I just try to profit from your lsp/completion setup, so I have less to manage myself.

Basically I try to convert my config and to upstream whatever seems useful.

PR #54, opened 2022-02-12
Do not set up mappings for plugins that were disabled

hunger:


hunger:
Latest version has the stray print removed.

PR #56, opened 2022-02-12
Allow to register extra snippets from vscode

hunger:
Allow users to register their own set of VSCode snippets from a set of
paths.


hunger:
This should be pre-commit clean now.

Issue #69, opened 2022-02-15
Move Bufferline theming to highlights


mehalter:
Other themes theme Bufferline using highlights. We should adopt this to make it easier to adopt other themes.

Here's an example from another theme: https://github.com/catppuccin/nvim/blob/main/lua/catppuccin/core/integrations/bufferline.lua

mehalter:
To go in `lua/onedark/others.lua`

```lua

  -- Bufferline
  BufferLineFill = { fg = C.grey_1, bg = C.black_1 },
  BufferLineBackground = { fg = C.grey_1, bg = C.black_1 },
  BufferLineBufferVisible = { fg = C.grey, bg = C.black },
  BufferLineBufferSelected = { fg = C.grey, bg = C.black, style = C.none },
  BufferLineTab = { fg = C.grey, bg = C.black },
  BufferLineTabSelected = { fg = C.black, bg = C.black },
  BufferLineTabClose = { fg = C.black, bg = C.black },
  BufferLineIndicatorSelected = { fg = C.black, bg = C.black },
  BufferLineSeparator = { fg = C.black_1, bg = C.black_1 },
  BufferLineSeparatorVisible = { fg = C.black, bg = C.black },
  BufferLineSeparatorSelected = { fg = C.black_1, bg = C.black_1 },
  BufferLineCloseButton = { fg = C.grey_1, bg = C.black_1 },
  BufferLineCloseButtonVisible = { fg = C.grey_2, bg = C.black },
  BufferLineCloseButtonSelected = { fg = C.red, bg = C.black },
  BufferLineModified = { fg = C.red, bg = C.black_1 },
  BufferLineModifiedVisible = { fg = C.grey, bg = C.black },
  BufferLineModifiedSelected = { fg = C.green, bg = C.black },
```

PR #75, opened 2022-02-16
Moved bufferline styling to highlight groups

mehalter:
Fixes: #69


mehalter:
@kabinspace I migrated the original colors over to the color table for the theme and updated the highlight group definitions. This should cause no change in colors from not using the color groups

PR #78, opened 2022-02-16
Remove Outline from Bufferline

mehalter:
When the Symbol Outline is part of the Bufferline it causes weird behavior when opening new files and stuff from the file browser when the outline is open and also makes an empty tab for the symbols. Seems to make more sense to ignore it.


mehalter:
Huh, not sure why the pre-commit.ci check is failing :(

PR #81, opened 2022-02-17
Allow overriding which-key options

Tilley:


kabinspace:
Very well, thank you for you contribution :)

PR #93, opened 2022-02-21
Checking the OS to correctly show how many pluing are loaded

Aryevang:
Will provide the correct count of loaded plugins when the using Neovim on Windows.

### DIFF
--- a/lua/configs/dashboard.lua
+++ b/lua/configs/dashboard.lua
@@ -6,6 +6,10 @@ function M.config()
 
   local plugins_count = fn.len(fn.globpath("~/.local/share/nvim/site/pack/packer/start", "*", 0, 1))
 
+  if fn.has("Win32") == 1 then
+    plugins_count = vim.fn.len(vim.fn.globpath("~/AppData/Local/nvim-data/site/pack/packer/start", "*", 0, 1))
+  end
+
   g.dashboard_disable_statusline = 1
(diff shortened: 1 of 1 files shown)


kabinspace:
@Aryevang Can you put the linux [plugin_count](https://github.com/kabinspace/AstroVim/blob/49d8c8c3bdbb39ac144ecee5e24dc740fe70db8c/lua/configs/dashboard.lua#L7) in else statement after the windows plugin_count?

Aryevang:
Sure. Tomorrow I'll update the PR.

Aryevang:
@kabinspace Done. I tried to satisfy the pre-commit.ci process but is a tricky boy.

lua/configs/dashboard.lua
```diff
@@ -45,7 +47,7 @@ function M.config()
 
   g.dashboard_custom_footer = {
     " ",
-    " AstroVim loaded " .. plugins_count .. " plugins ",
+    " AstroVim loaded " .. plugin_count .. " plugin ",
```
kabinspace, line 50: @Aryevang can you make it `plugins` instead `plugin in the end? since there are multiple plugins it would be better if we use plural form :)

Aryevang: @kabinspace, sorry about that. That was a wild replace :)

PR #98, opened 2022-02-24
Do not apply dashboard autocmd without dashboard

hunger:
Do not medddle with dashboard settings without the dashboard!


hunger:
In fact this probably does only make sense when both the dashboard and the bufferline are enabled.

PR #117, opened 2022-03-04
Remove duplicate mappings

### LINKED ISSUE #110: Remove duplicate keybindings
There are some duplicate keybindings in mappings.lua with which-key.lua. This should be fixed.

mehalter:
Fixes #110


mehalter:
Need to rewrite this. We want to leave the bindings in  `mappings.lua` just in case someone disables `which-key`. To resolve the duplication, I am going to migrate all of the bindings to `mappings.lua` and instead just use the `which-key.lua` file to put documentation for them.

PR #118, opened 2022-03-04
Allow to polish plugins before loading

### LINKED ISSUE #113: Allow user control of lazy loading of plugins
Plugins like ToggleTerm and Telescope for example are plugins that I have needed to disable lazy loading. It would be good to have a generic way to do this just in case someone else wants to disable lazy loading of various plugins

### LINKED ISSUE #114: Make more plugins optional
It would be nice to make more plugins optional. I personally need to disable autopairs and figured it would be good if we make nearly all plugins optional as a good bit of them are

hunger:
This change allows users to manipulate the lua table passed to packer to
load plugins, allowing to override every aspect of the configuration.

Note that this is very low level and not suited to change plugin
configuration since there the configuration suggested by AstroVim is
lost at this point.

Fixes #113, Fixes #114

See the README.md change for an example on how to use this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,25 @@ Just copy the `packer` configuration without the `use` and with a `,` after the 
 
 See the example above.
 
+### Change Configfuration of default Plugins
+
+AstroVim provides a `polish_plugins` function in the user settings that can be used to override the packer
+configuration for all plugins, user plugins as well as plugins configured by AstroVim.
+
+E.g. this code in your `settings.lua` will globally disable the lazy loading that is used by AstroVim by default:
+
+```lua
+  polish_plugins = function(plugins)
+    local result = {}
+    for _, plugin in pairs(plugins) do
+      plugin["cmd"] = nil
+      plugin["event"] = nil
+      table.insert(result, plugin)
+    end
+    return result
+  end
+```
+
 ### Adding sources to `nvim-cmp`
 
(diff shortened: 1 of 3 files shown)


hunger:
I think this is just a problem in the Ci script:-( Unfortunately I can not just rerun the check.

mehalter:
@hunger, Could you add an example to the user config? Even just a commented out example would make it easier for new users to know how to use it. Also does the polish only disable the lazy loading or does it allow extensibility of any part of the plugin config?

hunger:
You get the list of all plugins that astrovim wants to load and can do anything you want with it. Remove plugins? Just filter them out. Change plugin config? Just replace the config function. Add a plugin? Just append it.

This is a full-power but also very low-level tool: Adding a plugin can be done much easier after all:-) The polish_plugins override should definitely not replace that! We also should have a nice way to override plugin configuration in addition to this.

But this is the full-power escape hatch to do everything imaginable to the plugins -- probably in a way that is not ever going to be too convenient to use. Upside: It enables stuff like disabling lazy loading globally. Nobody in there right mind will want to do that, but it is great that we can support even crazy things like that with the necessary hacks in the user config -- keeping these hacks out of astrovim itself!

hunger:
I added a bit about `polish_plugins` to the Extending section in the README.

mehalter:
Fixes #114, #113

hunger:
The check was OK before I updated the commit message:-/

PR #120, opened 2022-03-05
Make cmp more predictable and usable

mehalter:
This commit does a few things to improve the usability of `cmp`

1. It does not auto select the first entry in the list. It must be specifically selected with ctrl-j/k or arrow keys. At the end of lines if I just want to type a new line and press enter and there is a possible selection it inserts it instead of a new line even though I did not select it.
2. Disabled the tab/s-tab for moving up and down in the list of completions. This is already enabled with ctrl-j/k and arrow keys and while the completion menu is open you can no longer jump between locations in a snippet without first pressing ctrl-e to close the menu.

Both of these issues cause behavior to change based on the menu being open and closed when interacting with the snippets and requires the user to think more before pressing keystrokes and whether or not they need to press `ctrl-e` first to close the menu.

### DIFF
--- a/lua/configs/cmp.lua
+++ b/lua/configs/cmp.lua
@@ -11,11 +11,6 @@ function M.config()
     return
   end
 
-  local check_backspace = function()
-    local col = vim.fn.col "." - 1
-    return col == 0 or vim.fn.getline("."):sub(col, col):match "%s"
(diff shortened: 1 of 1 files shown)


hunger (approved):
I like this!

PR #123, opened 2022-03-06
Add arbitrary user configuration of default plugins

### LINKED ISSUE #111: Configurable sidebar widths
Allow the user to configure the default width of NvimTree and SymbolsOutline

### LINKED ISSUE #112: User customizable Telescope config table
Allow the user to configure the telescope config table with changes to keybindings and extension configuration. Basically for extensions that need more than just loading the extension. For example `telescope-media-files` and `telescope-hop`

### LINKED ISSUE #115: User configurable cmp bindings
Allow the user to edit cmp config table (such as keybindings)

mehalter:
This sets up a system that extends the `overrides` table in the user settings where the user can arbitrarily overwrite and extend any default plugin configuration files. 

Fixes #115, Fixes #112, Fixes #111

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,21 +21,28 @@ AstroVim is an aesthetic and feature-rich neovim config that is extensible and e
 </p>
 
 ## 🌟 Preview
+
 ![Preview1](./utils/media/preview1.png)
 ![Preview2](./utils/media/preview2.png)
 ![Preview33](./utils/media/preview3.png)
 
 ## ⚡ Requirements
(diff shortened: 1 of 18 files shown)


hunger:
Why did you opt for requireing files for the config?

I would prefer if we could keep the overrides section in the settings.lua file: That would keep treesitter config compatible (I think). You can easily delegate to user-defined files by doing somehing like

```lua
overrides = {
   foo = require "user.foo",
}
```

mehalter:
@hunger this is a good point. I'll make it pull from the single file

PR #150, opened 2022-03-14
Clean up default user config

mehalter:
This makes the `user/` directory optional when no user customization is needed. The user must now create their own `user/` directory and we provide `user_example/` as a template. This lets the user's custom settings be truly untracked and not lead to merge conflicts on new updates.


abalmos:
For what it is worth (which is likely not much), I see this as a negative change.

Currently, I either get the new settings in a clean merge (which is easily reviewed and tweaked if I choose) or a conflict which positively tells me that I need to resolve something major. 

With this change, I have manually hunt through git/change logs. I moved from NvChad for several reasons, this being one of the smaller reasons.

mehalter:
This is great feedback. I hadn't thought of that. I can remove the change to the user directories and just keep the failsafe for the lack of a user defined `null-ls.lua` file

mehalter:
@abalmos the way this set up works with the `user/` folder is it only provides overrides for the default settings so any new settings added will be respected over time given your own user folder does not override them so you wouldn't need to worry about missing out on new settings getting set. Currently several users are having issues with merge conflicts breaking the update path and causing a workflow where they need to move their own settings away, `git pull`, and move them back.

But this PR would make sure that any new changes that are added to the default set up do get populated to the user in the end on updates without any necessary editing.

abalmos:
@mehalter That is fair. However, as an example, I didn't learn about the `polish` function until after it was added to the default `user/settings.lua` file and it caused a merge conflict. Until then I just wondered (without time to investigate) why my mappings stopped working. I don't think this is game changing, though. I just need to keep a closer on when updating.

PR #165, opened 2022-03-19
Make telescope-fzf more stand-alone

hunger:
The extension is the default anyway, so I just removed that...

This makes it possible to just disable the fzf plugin without killing the telescope plugin as well -- e.g. because you can not actually build the fzf binary this needs:-)

### DIFF
--- a/lua/configs/telescope.lua
+++ b/lua/configs/telescope.lua
@@ -7,7 +7,6 @@ function M.config()
   end
 
   local actions = require "telescope.actions"
-  telescope.load_extension "fzf"
 
   telescope.setup(require("core.utils").user_plugin_opts("plugins.telescope", {
     defaults = {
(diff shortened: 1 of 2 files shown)


mehalter (approved):
LGTM

AXGKl:
Hi, just tried astrovim on a clean server, worked, liked it. 

=> installed on my laptop, got this error: 





back to the server, git pull (got this PR now, was merged 2 hours before) => same error


I.e. the error is definitely caused by this PR, git checkout from before and it worked again.

mehalter:
@AXGKl  This should be fixed with #167 sorry for the issue!

AXGKl:
for what it's worth: #167 fixes it for me as well, on both machines.

mehalter:
Ok thanks! Hopefully when @kabinspace is around he can approve and merge my PR :)

PR #173, opened 2022-03-21
Add lazy-loading back to friendly-snippets

bryant-the-coder:
Fix lazy-loading snippets without breaking them like before #172


mehalter:
This does not work, `friendly-snippets` does not load properly.

PR #193, opened 2022-03-23
remove conflicting rn binding

mehalter:
Renaming is taken care of with the `<leader>lr` binding and `rn` conflicts with replacing a character with `n`


abalmos:
Would `<leader>rn` be more natural?

mehalter:
@abalmos I think that would be a good option. I think the process we were using is we want the `which-key` menu to be nicely organized especially for people learning the new bindings through `which-key`. Where `<leader>l` covers all of the LSP functionality. I will leave this up to @kabinspace since it's his project and anything that is set by default can always be changed in my own user files if it is not generally agreed on :)

kabinspace (approved):

mehalter:
Another option is we can keep the `<leader>lr` for a which-key menu item and just have `<leader>rn` work anyway. @abalmos and it just be listed on the wiki under default bindings.

kabinspace (approved):

PR #194, opened 2022-03-23
remove conflicting ca binding

mehalter:
Code action is taken care of with `<leader>la` and this conflicts with a very important part of the vim grammar. `ca<object>` for changing something like `ca[` for changing all of a `[]` surrounding area.


mehalter:
@abalmos what do you think of this binding as well just to get as many eyes on default changes as possible :)

kabinspace (approved):

PR #195, opened 2022-03-23
Move from LSP Saga to Dressing

mehalter:
LSP Saga project has been pretty inactive. This maintains the pretty look of UI elements while also using telescope for the default `vim.ui.select`.

Also by switching we are getting some decent startup time improvements.


hunger:
I use this as a basis for my config now and it seems to work fine.

One annoyance is that the rename window pops up over the item I want to rename. Apart from that the UI changed a bit in places, but seems to be just as usable.

hunger:
Startup time is below 100ms on my machine with this change (when removing everything that might trigger packer to lazy load anything;-).

hunger (approved):

kabinspace (approved):

PR #222, opened 2022-03-29
Add basic DAP support

hunger:
DAP is awfully handy to debug some things, especially with language support extensions that hook into this generic DAP support (e.g. rust-tools.nvim).

With this configuration + rust-tools I can hit "K" on a main function, select "debug" in the hover popup and have the DAP UI come up after a while;-)

### DIFF
--- a/lua/configs/dap.lua
+++ b/lua/configs/dap.lua
@@ -1,0 +1,32 @@
+local M = {}
+
+function M.ui_config()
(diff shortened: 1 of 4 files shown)


hunger:
I used the "enabled" config section for this since this setting is supposed to cover a range of plugins and keymaps, so that seemed appropriate.

PR #234, opened 2022-03-30
Add tab selection for cmp

### LINKED ISSUE #230: Having issues in select mode.
In select mode,(as first charecter) pressing p is pasting from clipboard(like visual mode) and pressing space is not inserting any space. Can you please check this problem? It's not letting use snippets properly.

kabinspace:
Adds back tab completion to CMP and fixed `p` when navigating snippets

Fixes #230

### DIFF
--- a/lua/configs/cmp.lua
+++ b/lua/configs/cmp.lua
@@ -11,11 +11,6 @@ function M.config()
     return
   end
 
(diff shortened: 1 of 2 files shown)


mehalter (approved):
LGTM

PR #243, opened 2022-03-31
Make lualine follow theme colors

mehalter:
This PR is going to make lualine play more nicely with other themes as well as make it significantly easier for the user to config from the ground up in `user/init.lua` instead of being forced into the default.

### DIFF
--- a/lua/configs/lualine.lua
+++ b/lua/configs/lualine.lua
@@ -2,26 +2,25 @@ local M = {}
 
 local status = require "core.status"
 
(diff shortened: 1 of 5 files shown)


mehalter:
This is very close, there are a few things left to do with this:

1. Make an actual default lualine theme instead of colors in the lualine config
2. Move the sections to the actual lualine config instead of `ins_left`/`ins_right`/etc.

kabinspace (approved):

PR #273, opened 2022-04-10
Remove duplicated diagnostic sign definition (LSP)

maxime50:
I just discovered that the diagnostic signs are already defined here:

https://github.com/kabinspace/AstroVim/blob/187c756f7f1c5168e3ba88a2ce55f71837378ab6/lua/configs/lsp/handlers.lua#L4-L9

So I removed those from the Neo-tree config as we don't really need to redefine them again.

Also corrected a small typo in the filetype config. 🙂

### DIFF
--- a/lua/configs/filetype.lua
+++ b/lua/configs/filetype.lua
@@ -1,7 +1,7 @@
 local M = {}
 
 function M.config()
-  local present, better_escape = pcall(require, "filetype")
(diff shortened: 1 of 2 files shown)


mehalter:
Thanks for the contribution @maxime50 !

mehalter (approved):
LGTM

PR #291, opened 2022-04-15
cmp: Use priorities to order sources

hunger:
This makes it easier to add custom sources in the right place. Just add
this into the config section of the plugin that implements the `foobar`
source:

```lua
  local cmp = require "cmp"
  local config = cmp.get_config()
  table.insert(config.sources, { name = "foobar", priority=1100 })
  cmp.setup(config)
```

This will make the `foobar` source show up at the top of the list, while a
priority of 100 will place it at the very bottom.

This avoids having to set the entire list of sources in one place (and
in the right order). That is not very composable. Check popular configs
based on astrovim and you will see this anti-pattern:-)

### DIFF
--- a/lua/configs/cmp.lua
+++ b/lua/configs/cmp.lua
@@ -77,10 +77,10 @@ function M.config()
       keyword_length = 1,
     },
     sources = {
-      { name = "nvim_lsp" },
-      { name = "luasnip" },
-      { name = "buffer" },
-      { name = "path" },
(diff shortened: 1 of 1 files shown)


mehalter:
I agree with this, I have set up a separate PR that implements this slightly differently but utilizes the same ideas. Thanks for the suggestion! I knew there had to be a better way of doing this! Development moved to #292

PR #321, opened 2022-04-20
Removed a lot of unnecessary redefinition of defaults and setup cleanup

mehalter:
We are redefining a lot of defaults unnecessarily. It makes it difficult to see exactly what AstroNvim is changing from the defaults of plugins. This does a lot of cleanup removing unnecessary lines from our config without changing the end product. Also this unifies all of our `X.setup()` calls to look the same across `config/X.lua` files.


mehalter:
This also decreases our chance of errors coming up as plugin updates happen. If a plugin changes around where their setup options are and we are relying on defaults, those changes might be taken care of by the plugin instead of us needing to keep track and updating our plugin configurations.

kabinspace (approved):

PR #328, opened 2022-04-21
Detail lualine and status

JamsMendez:
Small detail in lualine and the status.

I am using windows terminal and it happens to me with VictorMono, FiraCode CaskaydiaCode

Screenshot 2022-04-21 155607

### DIFF
--- a/lua/configs/lualine.lua
+++ b/lua/configs/lualine.lua
@@ -64,7 +64,7 @@ function M.config()
           },
           {
             "diff",
-            symbols = { added = " ", modified = "柳", removed = " " },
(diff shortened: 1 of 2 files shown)


mehalter (approved):
LGTM

mehalter:
Ah thanks for catching this!

kabinspace:
@JamsMendez we are going to revert this change as the extra padding added from the PR made the lualine's component appear too far making it look odd. I believe the problem is from your side since some terminals have bad font support for unicode characters, I would recommend you to change your terminal to get the proper support. Also, you can change those paddings from `nvim/lua/user/init.lua` if you like.

PR #348, opened 2022-04-25
Add ability for easy lsp setup without lsp installer

### LINKED ISSUE #343: Adding already installed LSP to lspconfig required list
Hi everyone,

Sorry if this was a dumb question, but I couldn't find anything in the documentation about it. I already have pyright and clangd installed on my system and I just wanted to add this to the cmp configuration like in their documentation:

require('lspconfig')['<YOUR_LSP_SERVER>'].setup {
    capabilities = capabilities
  }
  
Is there a way to do this? If I install through :LSPInstall will it override my current versions?
  
Thank you in advance!

mehalter:
Resolves #343

This adds the ability for easily setting up language servers without LSP installer either because LSP installer doesn't support installing the server or if the user just already has the server installed on their machine. Here is an example `user/init.lua` for adding `pyright` to the LSP config where the user already has `pyright` installed and doesn't want to use `lsp-installer`:

```lua
return {
  lsp = {
    setup = function(on_attach, capabilities)
      require("lspconfig").pyright.setup {
        on_attach = on_attach,
        capabilities = capabilities,
      }
    end,
  },
}
```

### DIFF
--- a/lua/configs/lsp/init.lua
+++ b/lua/configs/lsp/init.lua
@@ -1,5 +1,10 @@
 local status_ok, _ = pcall(require, "lspconfig")
 if status_ok then
   require "configs.lsp.lsp-installer"
-  require("configs.lsp.handlers").setup()
+  local handlers = require "configs.lsp.handlers"
+  handlers.setup()
+  local lsp_setup = require("core.utils").user_plugin_opts "lsp.setup"
+  if type(lsp_setup) == "function" then
+    lsp_setup(handlers.on_attach, handlers.capabilities)
+  end
 end


kabinspace (approved):

mehalter:
Actually, upon further investigation I have realized this is already easily done in the `polish` function of a `user/init.lua` file:

```lua
return {
  polish = function()
    local lsp_avail, lspconfig = pcall(require, "lspconfig")
    if lsp_avail then
      local handlers = require "configs.lsp.handlers"
      lspconfig.pyright.setup {
        on_attach = handlers.on_attach,
        capabilities = handlers.capabilities,
      }
    end
  end,
}
```

PR #364, opened 2022-04-27
Replace lazygit with gitui

krishnakumarg1984:
[GitUI](https://github.com/extrawurst/gitui) is a TUI for git that appears  to be a better (IMO) replacement for lazygit in terms of functionality and performance (see benchmark comparisons of gitui vs lazygit in the project's readme).

This PR replaces lazygit with gitui.  GitUI is 
- MIT licensed
- Written in rust, is cross-platform, and has unit test and CI pipelines
- easily installable by the user via multiple methods (package managers such as cargo, brew, nix, scoop, chocolatey etc. or can use one of the pre-built binaries provided for many platforms).

This PR replaces `lazygit` with `gitui` in AstroNvim. One of the toggle-term key binding is changed to `<leader>tg` for consistent mnemonics. I have tested `gitui` on various large repos and didn't find any issues. The maintainers of AstroNVim (and other community users) could perhaps do further testing on their own git projects before considering merging this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ git -C ~/.config/nvim remote set-url origin https://github.com/AstroNvim/AstroNv
 - Terminal with true color support (for the default theme, otherwise it is dependent on the theme you are using)
 - Optional Requirements:
   - [ripgrep](https://github.com/BurntSushi/ripgrep) - live grep telescope search (`<leader>fw`)
-  - [lazygit](https://github.com/jesseduffield/lazygit) - git ui toggle terminal (`<leader>tl` or `<leader>gg`)
+  - [gitui](https://github.com/extrawurst/gitui) - git ui toggle terminal (`<leader>tg` or `<leader>gg`)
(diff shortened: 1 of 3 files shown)


mehalter:
We'll look into this to see if this is a good fit and test it. In the mean time you can make this change to your user config without making a PR if you want

mehalter:
I think for my own personal use case I prefer `lazygit` to GitUI. If you want to achieve this functionality in your `init.lua` file I would recommend doing something like the following:

```lua
return {
  ["which-key"] = {
    register_mappings = {
      n = {
        ["<leader>"] = {
          g = {
            g = {
              function()
                require("core.utils").toggle_term_cmd "gitui"
              end,
              "GitUI",
            },
          },
          t = {
            l = { nil },
            g = {
              function()
                require("core.utils").toggle_term_cmd "gitui"
              end,
              "GitUI",
            },
          },
        },
      },
    },
  },
}
```

krishnakumarg1984:
Thank you. For now, I shall just use the user-only `init.lua` configuration suggested by you.

PR #376, opened 2022-04-28
Remove all old lsp-installer APIs

mehalter:
`nvim-lsp-installer` is planning on deprecating some old APIs and decouple itself more form `lspconfig` and configuring LSP servers. This adopts the new approach that they are moving to (our old approach will be going away). This also makes it easier and more straight forward to add language servers for servers that are already installed and aren't using `nvim-lsp-installer`.


mehalter:
This PR has a couple different approaches across a couple different commits. I asked a question as to how far the deprecation is going to do: https://github.com/williamboman/nvim-lsp-installer/pull/631 depending on this answer we will select the appropriate approach for the new LSP configuration setup.

PR #378, opened 2022-04-28
Add LSP Extension Plugin Compatibility

mehalter:
If a plugin sets up LSP specific capabilities and on_attach functions then we should respect that when we call our own language server setups.


kabinspace (approved):

hunger (changes requested):

hunger (approved):
👍🏼

PR #386, opened 2022-04-29
Move from Symbols Outline to Aerial

### LINKED ISSUE #377: Use aerial.nvim as the symbols viewer (instead of symbols-outline.nvim) due to ability to use treesitter for symbols extraction
symbols-outline.nvim requires a properly configured LSP server & attached to the current buffer. 

- Installing an LSP is not always possible, especially when working on remote machines without admin access. 
- As a workaround for the lack of sudo access, it can be argued that userspace installs are possible (with brew, nvim-lspinstall, compile from source etc), there are several problems in doing this:
1. A lot of LSPs require nodejs as a dependency (which may not be available or needs to be built!)
2. Even if we get past that, some LSPs require a very modern runtime environment. For example, clangd requires a fairly recent version of `glibc`. In machines that still run CentOS 7 (which is still supported) such as in the HPCs I use, administrators of such national supercomputing facilties shall not update these core infrastructure packages.

We need a plugin that does not rely on LSP to provide a symbols-tree outline viewer. 

Use a package that can leverage `treesitter` for providing symbols-tree viewer. A plugin that I have been using for the last 5 months is [aerial.nvim](https://github.com/stevearc/aerial.nvim), and I have been satisfied with it. It is actively maintained, and the author is responsive to address questions & GitHub issues. 

Quoting the relevant line of text from the project's README:
>First ensure you have [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter) installed and configured for all languages you want to support. That's all! Aerial will automatically fetch symbols from treesitter.

On remote machines without admin access, there are no impediments to using treesitter, since there is built-in support in installing the relevant parsers in core `neovim` itself (the latest version of nvim can be used as an appimage downloaded to my `$HOME/bin`).

mehalter:
This is a work in progress PR for moving to Aerial from Symbols Outline for more general support with Treesitter.

# TODO

- [x] investigate padding to make it look nicer (not currently an option)
- [x] figure out what symbols we want to show up since Symbols Outline assumes all we can make this better
- [x] add highlight groups
- [x] profit

Resolves #377

### DIFF
--- a/lua/configs/aerial.lua
+++ b/lua/configs/aerial.lua
@@ -1,0 +1,47 @@
+local M = {}
+
+function M.config()
+  local present, aerial = pcall(require, "aerial")
+  if present then
+    aerial.setup(require("core.utils").user_plugin_opts("plugins.aerial", {
+      close_behavior = "global",
+      backends = { "lsp", "treesitter", "markdown" },
+      min_width = 35,
+      show_guides = true,
+      filter_kind = {
+        "Class",
+        "Constructor",
+        "Enum",
+        "Function",
+        "Interface",
+        "Module",
+        "Method",
+        "Struct",
+        "Variable",
+      },
+      icons = {
+        Array = " ",
+        Boolean = "⊨ ",
+        Class = "𝓒 ",
+        Key = "🔐 ",
+        Namespace = " ",
+        Null = "NULL ",
+        Number = "# ",
+        Object = "⦿ ",
+        Property = " ",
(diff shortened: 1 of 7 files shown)


mehalter:
# TODO

- [ ] investigate padding to make it look nicer
- [ ] figure out what symbols we want to show up since Symbols Outline assumes all we can make this better
- [ ] profit

kabinspace (approved):

PR #394, opened 2022-04-30
Move from Dashboard to Alpha

### LINKED ISSUE #388: Move to Alpha.nvim
Dashboard.nvim is not actively developed, is slow on start up and it is also written in VimScript instead of Lua.

We should move to [Alpha.nvim](https://github.com/goolord/alpha-nvim) as a more future proof solution and faster start up times.

**Additional context**
We will also need to update the autocommands for disabling the tabline and statusline when the "alpha" (dashboard) is open.

### LINKED ISSUE #395: SessionLoad does not restore buffers
**Checklist**
Please confirm you have run the following commands and restarted Neovim and are encountering the issue with the current latest version of AstroNvim
- [x ] `:PackerSync`
- [ x] `:AstroUpdate`
- [ x] restarted AstroNvim
- [ x] Neovim version >= 0.7.0

Problem: Session restores but buffers do not open. As  #394  exists maybe this is not worth exploring. 

To reproduce: 
1.Open a bunch of buffers,
2.`Leader` `Ss` 
3. restart nvim
4. `leader` `sl`

Notice no buffers were restored.

mehalter:
- [x] add https://github.com/Shatur/neovim-session-manager
- [x] investigate lsp attachment issue when switching from one session to another

Resolves #388, fixes #395


kabinspace (approved):

lua/core/plugins.lua
```diff
@@ -286,11 +286,10 @@ if packer_status_ok then
 
     -- Start screen
     {
-      "glepnir/dashboard-nvim",
+      "goolord/alpha-nvim",
       config = function()
-        require("configs.dashboard").config()
+        require("configs.alpha").config()
       end,
-      disable = not config.enabled.dashboard,
```
thieung, line 293: It would be nice if you can make alpha can be lazy-loaded like current dashboard.

mehalter: This can be done with the `init.lua`:

```lua
return {
  plugins = {
    init = {
      ["goolord/alpha-nvim"] = { disable = true },
    },
  },
}
```

PR #403, opened 2022-05-02
Lazy load aerial plugin

thieung:
Allow user disable/enable Aerial plugin via custom config


datamonsterr (commented):
This is not lazyloading :))

mehalter:
@thieung this is actually a deprecated method of disabling plugins now that we have better user overriding. You can do this as follows in your `init.lua`:

```lua
return {
  plugins = {
    init = {
      ["stevearc/aerial.nvim"] = { disable = true },
    },
  },
}
```

mehalter:
I should update the docs to display this.

Issue #391, opened 2022-04-30
Preparation for v0.8


mehalter:
This is just a placeholder for keeping up with changes that are coming with Neovim v0.8.

## TODO

- [x] `resolved_capabilities` deprecated (https://github.com/neovim/neovim/pull/17814)
- [x] `nvim_cmd` API addition (https://github.com/neovim/neovim/pull/18366)
- [x] Use `feline` to set up `winbar` for breadcrumbs in active window and filename for inactive window (https://github.com/neovim/neovim/pull/18562)
- [x] Investigate `cmdheight = 0` (https://github.com/neovim/neovim/pull/16251)
- [x] Remove deprecated `filetype.lua` options (https://github.com/neovim/neovim/pull/19216)
- [x] https://github.com/neovim/neovim/pull/19238

mehalter:
The latest commit on the [`neovim-0.8` branch](https://github.com/AstroNvim/AstroNvim/tree/neovim-0.8) adds some winbar goodies

![2022-05-25_17:23:22_screenshot](https://user-images.githubusercontent.com/1591837/170369748-d71dea0a-7205-4c5c-a061-06ac0fb09a23.png)

mehalter:
Updated the `winbar` commit to make it look a lot nicer. Anyone using the `neovim-0.8` branch should do a `git pull --rebase` to get the changes :)
![2022-05-26_12:12:21_screenshot](https://user-images.githubusercontent.com/1591837/170529170-e3fd167f-5b16-4839-ab1c-ad7200a257a8.png)

mehalter:
Added `cmdheight=0` as default
![2022-06-14_08:15:10_screenshot](https://user-images.githubusercontent.com/1591837/173577236-f4af188c-6c9d-4384-b106-53bfc6308b17.png)

flamendless:
Im looking for breadcrumbs functionality, my previous config has it but since latest config uses a new statusline my old config for it will be harder to setup. Can this be used now? how?

mehalter:
@flamendless Yeah it should be happening automatically out of the box in the winbar as shown in the screenshot with feline.

mehalter:
With this new migration to Heirline we are also able to disable the WinBar on a per buffer basis. So now sidebars don't have a blank line at the top 
![2022-06-28_10:36:08_screenshot](https://user-images.githubusercontent.com/1591837/176207459-be400d14-1f7d-4a14-a30e-ee12b51324f3.png)

[...]

PR #421, opened 2022-05-02
Correct some deprecation warnings

jakeperry:
client.resolved_capabilities has been deprecated this replaces them with the client.server_capabilities as noted in the warning.

### DIFF
--- a/lua/configs/lsp/handlers.lua
+++ b/lua/configs/lsp/handlers.lua
@@ -42,7 +42,7 @@ function M.setup()
 end
 
 local function lsp_highlight_document(client)
-  if client.resolved_capabilities.document_highlight then
+  if client.server_capabilities.document_highlight then
     vim.api.nvim_create_augroup("lsp_document_highlight", {})
     vim.api.nvim_create_autocmd("CursorHold", {
       group = "lsp_document_highlight",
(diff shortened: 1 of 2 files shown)


mehalter:
This has only happened on nightly right? @jakeperry We aren't going to be adding official support for Nvim 0.8 or nightly until it's stable and more available. Please check out #391 for tracking what will be updated at some point. (This is added to the roadmap)

mehalter:
Also these fields have changed names completely to things like `server_capabilities.documentFormatProvider` just so you are aware. These are documented in the PR that was merged: https://github.com/neovim/neovim/pull/17814

As the code is now that you proposed it will get rid of the deprecation warning, but it will break those features.

mehalter:
@jakeperry For the time being to help with this migration and getting it started early I have made a branch for this with the necessary updates: https://github.com/AstroNvim/AstroNvim/tree/neovim-0.8

Feel free to use this branch, but there may be bugs and stuff as new stuff is added/changed in nightly. I will try to keep it up to date with `main` as well.

auser:
What's the ETA on 0.8? I'm using `main` branch but still getting these errors :/

mehalter:
@auser we will not be releasing official support for neovim 0.8 until it is released in stable. So my guess is probably like ~3 months until then? These deprecation changes are breaking between the newly released 0.7 and the current nightly so we would need to drop support which we aren't going to do for the current stable Neovim. If you want to use nightly, feel free to try the `neovim-0.8` branch. It should be up to date with master. I can't make any promises it will be bug free and it currently relies on a very new version of nightly so you may have to update it.

PR #422, opened 2022-05-02
Preparation for Neovim v0.8 Release

mehalter:
This is a heavy WIP just to keep on top of breaking changes on nightly as they come up. This branch should be considered unstable and in development

This has a small breaking change for people who do customization to `feline`. Instead of just a `plugins.feline` table since there are 2 `setup` calls we now have `plugins.feline.statusline` and `plugins.feline.winbar` for configuring each of the two bars.

### DIFF
--- a/lua/configs/feline.lua
+++ b/lua/configs/feline.lua
@@ -1,13 +1,15 @@
-local status_ok, feline = pcall(require, "feline")
-if status_ok then
+local present, feline = pcall(require, "feline")
+if present then
   local C = require "default_theme.colors"
   local hl = require("core.status").hl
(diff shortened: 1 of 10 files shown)


bryant-the-coder:
@mehalter ur fast af 😂

famiu:
You can just use `vim.cmd` with a table argument instead of passing an empty opts table to `nvim_cmd` thanks to https://github.com/neovim/neovim/pull/18523

mehalter:
Oh awesome @famiu ! Thanks for letting me know!

mehalter:
Development moved to #715

PR #447, opened 2022-05-04
turn off cursorline for telescope prompt because it's useless and ugly.

datamonsterr:


mehalter:
We will fix this elsewhere

PR #448, opened 2022-05-04
Split user_example config

Nawot:
I think be better right away show for user that he can split yourself config

### DIFF
--- a/lua/user_example/init.lua
+++ b/lua/user_example/init.lua
@@ -35,16 +35,6 @@ local config = {
     telescope_select = true,
   },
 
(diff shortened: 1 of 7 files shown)


mehalter:
@Nawot I think we have noticed that most users prefer to keep their edits in a single file `init.lua` rather than a bunch of separated files because they don't have very many edits and don't like having a bunch of small files with only a couple lines. For this reason I think we are going to prefer keeping the provided example relatively simple and just a single file and more advanced users can look at the documentation and browse the "black belt" configuration repos we link there as well for guidance on separating out the files.

PR #473, opened 2022-05-06
Add lazy loading to alpha

### LINKED ISSUE #471: Investigate lazy load of alpha-nvim
Currently `alpha-nvim` is always loaded, irrespective of whether a greeter screen is going to be shown or not.

The greeter screen is to be shown pnly when starting `nvim` by itself (i.e. without any files as its arguments). 

There is no need to always load `alpha-nvim` in other scenarios where a file is opened for editing, and  `alpha` can be loaded on demand when the user invokes it using `<leader>d` (or the `:Alpha` command). Can the project maintainers please look into lazy loading?

mehalter:
This takes a first stab at lazy loading alpha if neovim is opened with a file.

This lazy loading is not quite as sophisticated as the alpha starting conditions.

Resolves #471

### DIFF
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -28,6 +28,30 @@ if utils.is_available "alpha-nvim" then
     pattern = "alpha",
     command = "set laststatus=0 | autocmd BufUnload <buffer> set laststatus=3",
   })
+  cmd("VimEnter", {
+    desc = "Start Alpha when vim is opened with no arguments",
+    group = "alpha_settings",
+    callback = function()
+      -- optimized start check from https://github.com/goolord/alpha-nvim
+      local should_skip = false
+      if vim.fn.argc() > 0 or vim.fn.line2byte "$" ~= -1 or not vim.o.modifiable then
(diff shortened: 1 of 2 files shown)


krishnakumarg1984:
That doesn't look robust enough. This may fail with  `vim -u custom_rc` (and other command-line arguments).

Perhaps the best way to lazy load is to see if the buflist or arglist is empty, rather than parsing command-line args. I just tested the `:args` command, and it is empty for all these scenarios.

mehalter:
Yeah I agree, I think in the ideal way would be to lift the code from alpha itself

mehalter:
@krishnakumarg1984 check out the updated commit. I lifted the check from the Alpha plugin and did some optimization for our use case. This should be sufficiently sophisticated.

krishnakumarg1984:
Ah. That sure looks sophisticated. I am not a lua expert, but looks like that is handling the empty args list case, and its edge cases.

Plus, you've configured it as an "alpha" module. Nice!

kabinspace (approved):

PR #485, opened 2022-05-07
Enhancement: Start alpha dashboard when no existing buffers

thieung:
Add autocommand to start alpha dashboard when delete last buffer

### DIFF
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -13,6 +13,12 @@ cmd({ "VimEnter", "FileType", "BufEnter", "WinEnter" }, {
 })
 
 if utils.is_available "alpha-nvim" then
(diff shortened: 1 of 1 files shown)


mehalter (changes requested):
I like this idea, but the check is not sophisticated enough and can cause problems very easily. If I have a buffer open with a file and a new file in another buffer. I can edit that new file all I want but if I close the file I have open it will register as "no file left" and open alpha instead of giving me the file I have been editing. There are several ways to end up in the case where a buffername and filetype are not set but that doesn't mean that you have deleted your last buffer.

mehalter:
@thieung I changed the approach to solving this problem and opened a PR #488 . This should resolve the problems from this approach.

PR #487, opened 2022-05-07
fixed require for indent-line.lua

M3NIX:
At the moment installing AstroNvim gives following error:
astrovim-error

This PR should fix this.


mehalter (approved):
Thanks so much for this! LGTM

PR #488, opened 2022-05-07
If closing the last buffer, show alpha

mehalter:
Replaces #485

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -29,9 +29,13 @@ end
 
 -- Bufdelete
 if utils.is_available "bufdelete.nvim" then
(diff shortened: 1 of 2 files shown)


thieung:
Thanks @mehalter for your help to modify this, I will test this on my machine now.

thieung:
Tested, it works as my expected
cc @kabinspace

mehalter:
@thieung We have discussed this on the discord server and have decided not to make this default behavior in AstroNvim. I added a recipe to the documentation to make it easy for users to add this functionality to their user configuration. Please check that out here if you want to add this to your setup! https://astronvim.github.io/recipes/alpha#open-alpha-automatically-when-no-more-buffers

PR #493, opened 2022-05-08
Use Main branch for BufferLine and ToggleTerm

Numprt:
When installing the latest AstroNvim against a fresh installation of neovim 0.7 - there is an issue with the initial PackerSync.

Two plugins, BufferLine and ToggleTerm, have changed from using "master" to "main" as their default branch. This PR specifies to use the "main" branch for these two plugins.

### DIFF
--- a/lua/core/plugins.lua
+++ b/lua/core/plugins.lua
@@ -67,6 +67,7 @@ if packer_status_ok then
     -- Bufferline
     {
       "akinsho/bufferline.nvim",
+      branch = "main",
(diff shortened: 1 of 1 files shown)


mehalter:
This happens automatically. It picks the default branch. This developer recently renamed these branches from `master` to `main` so you need to delete and reinstall those plugins (annoying but out of our control sadly). Because Packer picks up the correct branch on the initial install we shouldn't be tracking these by name in the configuration.

PR #514, opened 2022-05-11
Lazy loading core plugins to save startup time

thieung:


mehalter:
Thanks for these suggestions. I'll test these out today to make sure they are good to go

lua/core/plugins.lua
```diff
@@ -76,6 +79,7 @@ if packer_status_ok then
     -- Better buffer closing
     {
       "famiu/bufdelete.nvim",
+      event = "BufReadPost",
```
mehalter, line 82: This shouldn't be here. This is lazy loaded on the commands so it's already lazy loaded more aggressively than this

lua/core/plugins.lua
```diff
@@ -307,6 +316,7 @@ if packer_status_ok then
       "akinsho/nvim-toggleterm.lua",
       cmd = "ToggleTerm",
       module = { "toggleterm", "toggleterm.terminal" },
+      event = "BufWinEnter",
```
mehalter, line 319: This is fully lazy loaded more aggressively. So this can be deleted

lua/core/plugins.lua
```diff
@@ -333,6 +346,7 @@ if packer_status_ok then
     {
       "folke/which-key.nvim",
       module = { "which-key" },
+      event = "VimEnter",
```
mehalter, line 349: This is also already lazy loaded on module. So this can be removed.

mehalter:
I'm going to clean this up in a dev branch and get this merged. Thanks @thieung for the contribution!

PR #518, opened 2022-05-11
BREAKING: Treat null-ls like every other plugin

mehalter:
Right now we have a very specific set up for `null-ls`. This makes it treated like every other plugin.


mehalter:
Going to merge this into a dev branch with Feline and we can merge the two together when the time comes

PR #523, opened 2022-05-12
chore: disable omnifunc

thieung:
[nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/wiki/Autocompletion#nvim-cmp) mentioned that we don't need to use omnifunc since  we are using `nvim-cmp`
> If you are using nvim-cmp do not use neovim's built-in omnifunc as it cannot support the additional completion items returned from servers due to the capabilities enabled by nvim-cmp.


mehalter (approved):
LGTM

PR #554, opened 2022-05-22
fix: call function setup instead of config for feline package

### LINKED ISSUE #555: Feline user config is ignored after update to the latest (791c4b2) commit
**Checklist**
Please confirm you have run the following commands and restarted Neovim and are encountering the issue with the current latest version of AstroNvim
- [x] `:PackerSync`
- [x] `:AstroUpdate`
- [x] restarted AstroNvim
- [x] Neovim version >= 0.7.0

**Describe the bug**
After upgrading to the latest feline version the status bar config defaults to the package config ignoring my customizations or AstroNvim defaults.

**To Reproduce**
Steps to reproduce the behaviour:
1. Sync packages
2. Restart nvim 

**Expected behavior**
The status bar should remain unchanged

**Screenshots**
If applicable, add screenshots or recording ([Asciinema](asciinema.org)) to help explain your problem.

**Additional context**

dragosboca:
Resolves #555 

After the latest update, the feline package requires calling the `setup` function, not `config` in the initialization section.

### DIFF
--- a/lua/core/plugins.lua
+++ b/lua/core/plugins.lua
@@ -82,7 +82,7 @@ local astro_plugins = {
   -- Statusline
   ["feline-nvim/feline.nvim"] = {
     after = "nvim-web-devicons",
-    config = function()
+    setup = function()
       require("configs.feline").config()
     end,
   },


mehalter:
This is not accurate. We should figure out why this is "fixing" this issue on your machine. The `setup` function for packer is run *before* the plugin is loaded so the `require("feline")` call should not work. We use `config` here because it is called after the plugin is loaded. I cannot replicate the issue you are having and changing this to `setup` on my machine breaks our feline configuration.

PR #559, opened 2022-05-25
fix(status): remove lsp_client_name bug in feline

datamonsterr:
I notice one problem that if you open a random file, the lsp_clients_name will display properly but if you open another file on another directory, lsp_clients_name won't exist on that file.

### DIFF
--- a/lua/core/status.lua
+++ b/lua/core/status.lua
@@ -78,7 +78,7 @@ end
 function M.provider.lsp_client_names(expand_null_ls)
   return function()
     local buf_client_names = {}
-    for _, client in ipairs(vim.lsp.buf_get_clients(0)) do
+    for _, client in pairs(vim.lsp.buf_get_clients(0)) do
       if client.name == "null-ls" and expand_null_ls then
         vim.list_extend(buf_client_names, astronvim.null_ls_sources(vim.bo.filetype, "FORMATTING"))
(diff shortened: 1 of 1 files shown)


lua/core/status.lua
```diff
@@ -78,15 +78,15 @@ end
...
+    for _, client in pairs(vim.lsp.buf_get_clients(0)) do
       if client.name == "null-ls" and expand_null_ls then
         vim.list_extend(buf_client_names, astronvim.null_ls_sources(vim.bo.filetype, "FORMATTING"))
         vim.list_extend(buf_client_names, astronvim.null_ls_sources(vim.bo.filetype, "DIAGNOSTICS"))
       else
         table.insert(buf_client_names, client.name)
       end
     end
-    return table.concat(buf_client_names, ", ")
+    return table.concat(buf_client_names, ", ") .. " "
```
mehalter, line 89: Could you please remove this change in this line as it is not necessary and messes up the padding.

mehalter (changes requested):
Thanks for this. I added a comment to remove one of the changes because it's not necessary and messes up the padding in the statusline. Also could you reword the commit to be `fix(status):` instead of `fix(core/status)`

lua/core/status.lua
datamonsterr: ok sure

datamonsterr: still think padding is better if someone want to have some background color on it

mehalter (approved):
LGTM

PR #562, opened 2022-05-26
chore(bufferline): make selected buffer noticeable in bufferline

mehalter:
Makes the currently selected buffer noticeable in the tab line. Useful when there are several visible buffers with splits.

2022-05-26_14:48:21_screenshot


krishnakumarg1984:
Does this work with any custom colorscheme that users have installed?

mehalter:
@krishnakumarg1984 no, this is just part of the default theme. It is up to the other custom themes to set the highlight groups for bufferline the way they want them

PR #623, opened 2022-06-11
config: Increase aerial disable_max_lines

uggla:
First, thank you AstroNvim is really cool.

This MR increases disable_max_lines to support symbols outline and search in large files.
Symbols search and outline were not possible with large python files, which is helpful with such kinds of files (that need to be split).
The default seems to be 10k lines, and this MR proposed to increase to 30k.
Performance is good even with files near 30k lines.

### DIFF
--- a/lua/configs/aerial.lua
+++ b/lua/configs/aerial.lua
@@ -6,6 +6,7 @@ if status_ok then
     min_width = 28,
     show_guides = true,
     filter_kind = false,
+    disable_max_lines = 30000,
     icons = {
       Array = "",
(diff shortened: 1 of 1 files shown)


mehalter:
I think we want to keep our plugins as close to default as we can especially when it comes to performance pertaining to older hardware which a lot of users do use. This can be overridden in your user configuration though

mehalter:
```lua
return {
  plugins = {
    aerial = {
      disable_max_lines = 30000
    }
  }
}
```

PR #625, opened 2022-06-11
feat: Remember cursor position

uggla:
This MR add a simple autocmd to remember the cursor position and keep it after closing and opening a file again.


mehalter:
This can be a tricky feature to implement for all use cases especially when folds and things are used so I'm not sure if we want to merge something like this upstream. I would recommend this plugin (it's what I have been using in my user config):


```lua
return {
  plugins = {
    init = {
      ["ethanholz/nvim-lastplace"] = {
        config = function()
          require ("nvim-lastplace").setup()
        end,
      }
    }
  }
}
```

PR #647, opened 2022-06-16
fix(core): Set colorscheme after opts

huss4in:
Some colorschemes which have multiple style options like:
- [marko-cerovac/material.nvim](https://github.com/marko-cerovac/material.nvim.git)
- [folke/tokyonight.nvim](https://github.com/folke/tokyonight.nvim)

Require their options to be set before setting up or enabling the colorscheme

---
So when I add the colorscheme plugin, and set it up like the following:
```lua
  -- Set colorscheme
| colorscheme = "tokyonight",

  -- set vim options here (vim.<first_key>.<second_key> =  value)
  options = {
    ...
    g = {
|     tokyonight_style = "night",
    },
  },

  ...

  -- Configure plugins
  plugins = {
    -- Add plugins, the packer syntax without the "use"
    init = {
|     { "folke/tokyonight.nvim" },
      ...
    },
    ...
},
```

Each time I open nvim the default style option is loaded, and I need to run `:colorscheme tokyonight` to load the `tokyonight_style = "night"` style option

---

Correct me if I'm doing it wrong or if I'm missing something, but I moved these lines which set up the colorscheme in `lua/core/options.lua` 

```lua
local colorscheme = astronvim.user_plugin_opts("colorscheme", nil, false)
vim.api.nvim_command(
  "colorscheme "
    .. (vim.tbl_contains(vim.fn.getcompletion("", "color"), colorscheme) and colorscheme or "default_theme")
)
```
to the end of the file, so the colorscheme is being set after the options

and this solves the issue

### DIFF
--- a/lua/core/options.lua
+++ b/lua/core/options.lua
@@ -1,9 +1,3 @@
-local colorscheme = astronvim.user_plugin_opts("colorscheme", nil, false)
-vim.api.nvim_command(
-  "colorscheme "
-    .. (vim.tbl_contains(vim.fn.getcompletion("", "color"), colorscheme) and colorscheme or "default_theme")
-)
-
 astronvim.vim_opts(astronvim.user_plugin_opts("options", {
   opt = {
     backspace = vim.opt.backspace + { "nostop" }, -- Don't stop backspace at insert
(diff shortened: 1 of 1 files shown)


huss4in:
I tried adding the plugin with the `setup` function as shown the the [documentation](https://astronvim.github.io/recipes/colorscheme#using-a-custom-colorscheme-configured-with-global-variables)

```lua
  -- Set colorscheme
  colorscheme = "tokyonight",

  -- set vim options here (vim.<first_key>.<second_key> =  value)
  options = {
    ...
    g = {
*     tokyonight_style = "night",
    },
  },

  ...
  plugins = {
    init = {
      {
        "folke/tokyonight.nvim",
|       setup = function()
|         vim.g.tokyonight_style = "night"
|       end,
      },
    },
  }
```

but this still didn't seem to load the `night` style of the theme when I open nvim,
I still need to run `:colorscheme tokyonight` to load it..

mehalter (approved):
LGTM

PR #690, opened 2022-06-22
docs: point to full list of supported clipboard tools

maxbrunet:
I spotted that in the latest release

`xclip` is for X11, `wl-clipboard` is for Wayland, and there actually are others:

```
                                                              clipboard-tool
The presence of a working clipboard tool implicitly enables the '+' and '*'
registers. Nvim looks for these clipboard tools, in order of priority:

  - g:clipboard
  - pbcopy, pbpaste (macOS)
  - wl-copy, wl-paste (if $WAYLAND_DISPLAY is set)
  - xclip (if $DISPLAY is set)
  - xsel (if $DISPLAY is set)
  - lemonade (for SSH) https://github.com/pocke/lemonade
  - doitclient (for SSH) http://www.chiark.greenend.org.uk/~sgtatham/doit/
  - win32yank (Windows)
  - termux (via termux-clipboard-set, termux-clipboard-set)
  - tmux (if $TMUX is set)
```

https://neovim.io/doc/user/provider.html#clipboard-tool

Windows does not seem to come with a pre-installed one either.

And as always, thank you for AstroNvim! :heart:

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -49,7 +49,7 @@ AstroNvim is an aesthetic and feature-rich neovim config that is extensible and 
 
 - [Nerd Fonts](https://www.nerdfonts.com/font-downloads)
 - [Neovim 0.7+](https://github.com/neovim/neovim/releases/tag/v0.7.0)
-- [xclip](https://github.com/astrand/xclip) - (Linux only) xclip is necessary for the clipboard integration with the system clipboard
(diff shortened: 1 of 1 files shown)


maxbrunet:
You might want to use `paths` to disable the above message when there is no config change
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

mehalter:
@maxbrunet thanks for this contribution! It will make the documentation a lot more precise and accurate :D

mehalter (approved):
LGTM

PR #742, opened 2022-07-09
feat: Add global API for setting highlight groups for any theme

mehalter:
Adds new `highlights` table to `user/init.lua` for setting hl groups for any theme:

```lua
return {
  highlights = {
    default_theme = {
      Normal = { bg = "#FFFFFF" },
    },
    duskfox = {
      Normal = { bg = "#000000" },
    },
    catppuccin = function(hl)
      local cp = require("catppuccin.api.colors").get_colors()

      hl.TSFuncBuiltin = { fg = cp.sapphire, bold = true }

      return hl
    end,
  },
}
```


mehalter:
Also lets you make a folder `user/highlights/` and any `.lua` file in it is the highlight table for the colorscheme with the name of the file. For example. `user/highlights/catppuccin.lua` could be used to set colors in Catppuccin colorscheme:

`user/highlights/catppuccin.lua`:

```lua
local cp = require("catppuccin.api.colors").get_colors()

return {
  TSFuncBuiltin = { fg = cp.sapphire, bold = true },
}
```

kabinspace (approved):

Johnsiras:
Hey @mehalter fmt key doesn't work in highlights do you know why?

PR #747, opened 2022-07-10
Update user_example config with new hl API example

pkazmier:
Add an example of the new global highlighting API to the `user_example` configuration for new users to discover.


mehalter (changes requested):
Thanks for the contribution! Maybe we should change the example to show the `default_theme` with the function notation so that users can see how to use the built in colors of the provided theme. Could you please edit your commit to do something similar to this? 

```lua
return {
  -- Override highlight groups in any theme
  highlights = {
    -- duskfox = { -- a table of overrides
    --   Normal = { bg = "#000000" },
    -- },
    default_theme = function(highlights) -- ... or a function that returns one
      local C = require "default_theme.colors"

      highlights.Normal = { fg = C.fg, bg = C.bg }
      return highlights
    end,
  },
}
```

mehalter:
Also @pkazmier when you edit your commit could you please reword it as well to match the Conventional Commit standard? Your commit message should be something like `chore(user_example): add new highlight api example`

pkazmier:
> Thanks for the contribution! Maybe we should change the example to show the `default_theme` with the function notation so that users can see how to use the built in colors of the provided theme. Could you please edit your commit to do something similar to this?
> 
> ```lua
> return {
>   -- Override highlight groups in any theme
>   highlights = {
>     -- duskfox = { -- a table of overrides
>     --   Normal = { bg = "#000000" },
>     -- },
>     default_theme = function(highlights) -- ... or a function that returns one
>       local C = require "default_theme.colors"
> 
>       highlights.Normal = { fg = C.fg, bg = C.bg }
>       return highlights
>     end,
>   },
> }
> ```

Sure, no problem. I actually had this originally, but then I figured I should use your example from the release notes instead.

pkazmier:
> Also @pkazmier when you edit your commit could you please reword it as well to match the Conventional Commit standard? Your commit message should be something like `chore(user_example): add new highlight api example`
> 
> For details on this, please check out the contributing guidelines: https://github.com/AstroNvim/AstroNvim/blob/main/.github/CONTRIBUTING.md

All set I think. Never did a force push before. Does it look correct?

mehalter:
Awesome! Looks great @pkazmier I'll review and approve it after the checks run.

PR #769, opened 2022-07-19
Add User experience menu

osamuaoki:
Building on discussion and tutoring at #764

I refactored original code set

### DIFF
--- a/lua/configs/which-key-register.lua
+++ b/lua/configs/which-key-register.lua
@@ -8,6 +8,7 @@ local mappings = {
       f = { name = "File" },
       p = { name = "Packer" },
       l = { name = "LSP" },
+      u = { name = "UI" },
     },
   },
 }
--- a/lua/core/utils/init.lua
+++ b/lua/core/utils/init.lua
@@ -258,5 +258,6 @@ function astronvim.cmd(cmd, show_error)
 end
 
 require "core.utils.updater"
+require "core.utils.ui"
 
 return astronvim
--- a/lua/configs/autopairs.lua
+++ b/lua/configs/autopairs.lua
@@ -21,6 +21,8 @@ npairs.setup(astronvim.user_plugin_opts("plugins.nvim-autopairs", {
   },
 }))
 
+if not vim.g.autopairs then npairs.disable() end
+
 local rules = astronvim.user_plugin_opts("nvim-autopairs").add_rules
 if vim.tbl_contains({ "function", "table" }, type(rules)) then
   npairs.add_rules(type(rules) == "function" and rules(npairs) or rules)
--- a/lua/configs/cmp.lua
+++ b/lua/configs/cmp.lua
@@ -63,6 +63,7 @@ setup(astronvim.user_plugin_opts("plugins.cmp", {
       border = { "╭", "─", "╮", "│", "╯", "─", "╰", "│" },
     },
   },
+  completion = { autocomplete = astronvim.set_cmp_autocomplete() },
   mapping = {
     ["<Up>"] = cmp.mapping.select_prev_item(),
     ["<Down>"] = cmp.mapping.select_next_item(),
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -9,7 +9,6 @@ maps[""]["<Space>"] = "<Nop>"
 maps.n["<leader>w"] = { "<cmd>w<cr>", desc = "Save" }
 maps.n["<leader>q"] = { "<cmd>q<cr>", desc = "Quit" }
 maps.n["<leader>h"] = { "<cmd>nohlsearch<cr>", desc = "No Highlight" }
-maps.n["<leader>u"] = { function() astronvim.toggle_url_match() end, desc = "Toggle URL Highlights" }
 maps.n["<leader>fn"] = { "<cmd>enew<cr>", desc = "New File" }
 maps.n["gx"] = { function() astronvim.url_opener() end, desc = "Open the file under cursor with system app" }
 maps.n["<C-s>"] = { "<cmd>w!<cr>", desc = "Force write" }
@@ -190,4 +189,21 @@ maps.t["<C-j>"] = { "<c-\\><c-n><c-w>j", desc = "Terminal down window navigation
 maps.t["<C-k>"] = { "<c-\\><c-n><c-w>k", desc = "Terminal up window navigation" }
 maps.t["<C-l>"] = { "<c-\\><c-n><c-w>l", desc = "Terminal right window naviation" }
 
+-- Custom menu for modification of the user experience
+if is_available "nvim-autopairs" then
+  maps.n["<leader>ua"] = { function() astronvim.toggle_autopairs() end, desc = "Toggle autopairs" }
+end
+maps.n["<leader>ub"] = { function() astronvim.toggle_background() end, desc = "Toggle background" }
+if is_available "nvim-cmp" then
+  maps.n["<leader>uc"] = { function() astronvim.toggle_cmp() end, desc = "Toggle completion" }
+end
+maps.n["<leader>ug"] = { function() astronvim.toggle_signcolumn() end, desc = "Toggle signcolumn" }
+maps.n["<leader>ui"] = { function() astronvim.set_indent() end, desc = "Change indent setting" }
+maps.n["<leader>un"] = { function() astronvim.change_number() end, desc = "Change line numbering" }
+maps.n["<leader>up"] = { function() astronvim.toggle_spell() end, desc = "Toggle spellcheck" }
+maps.n["<leader>uu"] = { function() astronvim.toggle_url_match() end, desc = "Toggle URL highlight" }
+maps.n["<leader>uw"] = { function() astronvim.toggle_wrap() end, desc = "Toggle wrap" }
+maps.n["<leader>uy"] = { function() astronvim.toggle_syntax() end, desc = "Toggle syntax highlight" }
+
 astronvim.set_mappings(astronvim.user_plugin_opts("mappings", maps))
+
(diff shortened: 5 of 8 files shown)


osamuaoki:
I realized my function definition style was not in line with other parts.  I will address it later.

osamuaoki:
* I only changed upper case usage style for" URL highligts" but wondering I should change to "URL highlight" without "s".
* I almost added "toggle paste mode" but since it is deprecated in nvim, I didn't.

lua/core/utils/ux.lua
```diff
@@ -0,0 +1,133 @@
...
+  elseif (number == true) and (relativenumber == false) then
+    vim.wo.number = false
+    vim.wo.relativenumber = true
+  elseif (number == false) and (relativenumber == true) then
+    vim.wo.number = true
+    vim.wo.relativenumber = true
+  else -- (number == true) and (relativenumber == true) then
+    vim.wo.number = false
+    vim.wo.relativenumber = false
+  end
```
mehalter, line 95: could we optimize and simplify this a little bit and do

```lua
  if not number and not relativenumber then
    vim.wo.number = true
  elseif number and not relativenumber then
    vim.wo.relativenumber = true
  elseif number and relativenumber then
    vim.wo.number = false
  elseif not number and relativenumber then
    vim.wo.relativenumber = false
  end
```

lua/configs/which-key-register.lua
```diff
@@ -8,6 +8,7 @@ local mappings = {
       f = { name = "File" },
       p = { name = "Packer" },
       l = { name = "LSP" },
+      u = { name = "User" },
```
mehalter, line 11: I think naming this to "User Interface" or even "UI" would be more clear what the menu options do.

lua/core/utils/init.lua
```diff
@@ -258,5 +258,6 @@ function astronvim.cmd(cmd, show_error)
 end
 
 require "core.utils.updater"
+require "core.utils.ux"
```
mehalter, line 261: could we rename this to `core.utils.ui` to match the "UI" menu option?

lua/core/utils/ux.lua
```diff
@@ -0,0 +1,133 @@
...
+  else
+    if vim.g.syntax_on then
+      vim.cmd("syntax off")
+    else
+      vim.cmd("syntax on")
+    end
+    local state = bool2str(vim.g.syntax_on)
+    print(string.format("syntax %s", state))
+  end
+end
```
mehalter, line 132: Could we simplify this to

```lua
function astronvim.toggle_syntax()
  local ts_avail, parsers = pcall(require, "nvim-treesitter.parsers")
  if vim.g.syntax_on then -- local to buffer
    if ts_avail and parsers.has_parser() then vim.cmd "TSBufDisable highlight" end
    vim.cmd "syntax off"
  else
    if ts_avail and parsers.has_parser() then vim.cmd "TSBufEnable highlight" end
    vim.cmd "syntax on"
  end
  local state = bool2str(vim.g.syntax_on)
  print(string.format("syntax %s", state))
end
```

mehalter (changes requested):
I made some comments with some optimizations/simplifications of the code. Also could we replace all of the `print` calls with `astronvim.notify` calls

lua/core/utils/ux.lua
pkazmier: What about adding a toggle wrap? Sometimes when editing markdown files, I notice the author hard wraps, other times they don’t. Would be nice to have a quick toggle. Thoughts?

pkazmier:
What about adding a toggle wrap? Sometimes when editing markdown files, I notice the author hard wraps, other times they don’t. Would be nice to have a quick toggle. Thoughts?

mehalter (changes requested):
I noticed you changed the prints to `astronvim.echo` could you change them to `astronvim.notify`?

osamuaoki:
> I noticed you changed the prints to `astronvim.echo` could you change them to `astronvim.notify`?
Woops.  s/echo/a...notify/

lua/core/utils/init.lua
osamuaoki: I thought about the same name but I didn't do so due to existing ui.lua file.
Now that I see you removed ui.lua, 
I will rename.
I will also drop tailing "s" from "Toggle URL highlights" to be consistent.

lua/core/utils/ux.lua
osamuaoki: Done with minor change

osamuaoki:
> What about adding a toggle wrap? Sometimes when editing markdown files, I notice the author hard wraps, other times they don’t. Would be nice to have a quick toggle. Thoughts?

Added ... (We can remove this if there is a objection.)

osamuaoki:
I enhanced this for nvim-autopairs.

osamuaoki:
I think this on-the-fly UI behavior change capability is getting matured now.

One shortcoming is if filetype plugins or something else sets indent settings, vales set by `<leader>ui` is ignored.  This happens with files like the lua source ones.  Considering this feature is some file not well supported by FTP such as plain text file, this is OK for now.  Any improvement idea is welcome.

Initial starting UI behavior customization should be a part of user_examples, I added them but I am 2 missing.
*  `vim.g.autopairs_toggle_flag`
*  `vim.g.cmp_toggle_flag`
If anyone knows best way, let me know.

lua/core/utils/ux.lua
osamuaoki: done

mehalter: Have you pushed this change? I don't see it

osamuaoki: Yes.  I moved it to ui.lua

mehalter: I meant the rewrite I recommended for the number setting cycling.

osamuaoki: FYI: I pushed to my ux branch which shows up on you account as osamu_ux
You may need to explicitly pull osamu_ux.
```
-- Change the number display modes.
function astronvim.change_number()
  local number = vim.wo.number          -- local to window
  local relativenumber = vim.wo.relativenumber  -- local to window
  if not number and not relativenumber then
    vim.wo.number = true
    vim.wo.relativenumber = false
  elseif number and not relativenumber then
    vim.wo.number = false
    vim.wo.relativenumber = true
  elseif not number and relativenumber then
    vim.wo.number = true
    vim.wo.relativenumber = true
  else -- number and relativenumber
    vim.wo.number = false
    vim.wo.relativenumber = false
  end
  vim.notify(
    string.format("number=%s, relativenumber=%s",
      bool2str(vim.wo.number),
      bool2str(vim.wo.relativenumber)
    )
  )
end
```

osamuaoki: Aha,…I see your point sir.

osamuaoki: Done this time.  Thanks

osamuaoki:
toggling for CMP and autopair needs some work.  (Otherwise initial mode setting doesn't worrk and it always start active.)

osamuaoki:
I think I need to work more on autopairs and cmp.  Excuse me.

osamuaoki:
I rebased to the current main.
  
DIFF: https://github.com/AstroNvim/AstroNvim/compare/main..e7aefed4a0463b08df5f55de4c9beb362a2af8ea

I think this is the best effort for me.

I could not figure out how to start autopairs with disabled state cleanly.
So it starts as ON which is current AstroNvim behavior.  (I tried autcmd etc...)
For cmp, I changed toggling method.

Thanks for interesting code base to play with.

osamu

mehalter:
Awesome, thanks so much @osamuaoki ! This is a great contribution and you have done a LOT of heavy lifting to get this PR to where it is :) I will take a look at what's currently there and merge it into a development branch to finish it up and get it merged. Thanks again :D

osamuaoki:
Thanks.

I think rebase conflict is minimal one since we both changed the last line of core/utils/init.lua.  Please make both changes accepted.

One thing I am concerned mildly is use of file name `ui.lua`.

Your development branch seems to have `core/ui.lua`.  I don't think this cause any real problem with `core/utils/ui.lua` in my code for lua-JIT but having the same file name may be confusing for humans.

osamuaoki:
I finally figured out how to start autopairs in disabled state.  It was packer setup .

Unlike, cmp, where we need a vim.g variable to toggle it, autopairs has internal state variable.  So in theory, we don't need to create vim.g variable and have fancy user configuration trick.

But that will be mind twisting to follow or write.  So I chose simple code.

So I can make all CMP/autopairs as opt-in feature.  I like them when I write first code but for debugging, it is annoying.  

(I was not able to get v2 running cleanly.  Do I need nvim 0.8 pre-release?)

mehalter:
Oh yeah @osamuaoki currently v2 does require neovim 0.8 nightly. Sorry about that!

osamuaoki:
> I noticed you changed the prints to `astronvim.echo` could you change them to `astronvim.notify`?

Done: This has been implemented.

I will squash changes and clean up ...

mehalter (changes requested):
@osamuaoki looks amazing! Thanks so much! Last thing though: could you reword your commit to say: `feat(mappings)!: add UI menu feature`

It's more of a mappings feature than a which-key one, also it does have a breaking change, so it needs the `!`

osamuaoki:
* I have updated commit message according to the request.
* I rebased this to the latest main.
* I added `if_available` to 2 mappings since user may disable the corresponding packages.  (I think this is right way.  Please double check modules.lua file.

Please note I have corresponding PR for web page. https://github.com/AstroNvim/astronvim.github.io/pull/13

In this PR, I also add "Configuration Mechanism" section so people can configure this package without reading the Lua source code around `user_plugin_opts`.  (Intentionally terse.)

mehalter (approved):
LGTM! I am going to merge this into a dev branch and do a quick rebase on the `v2` branch and merge it there since it has breaking changes. Thanks so so so much for your hard work on this one!

mehalter:
Finishing up at #816

osamuaoki:
Thanks.

PR #776, opened 2022-07-21
Fixes client.resolved_capabilities.document_formatting

pierrenel:
I kept on seeing this error message when opening .js files:

```resolved_capabilities is deprecated, update your plugins or configuration to access client.server_capabilities instead. The new key/value pairs in server_capabilities directly match those defined in the language server protocol```

via https://github.com/neovim/nvim-lspconfig/issues/1891 - editing handles.lua with these new values made the error go away. I also have no idea what I'm doing, but the error message is gone, so there's that :)

### DIFF
--- a/lua/configs/lsp/handlers.lua
+++ b/lua/configs/lsp/handlers.lua
@@ -31,7 +31,7 @@ astronvim.lsp.on_attach = function(client, bufnr)
     { desc = "Format file with LSP" }
   )
 
-  if client.resolved_capabilities.document_highlight then
+  if client.server_capabilities.DocumentHighlight then
     vim.api.nvim_create_augroup("lsp_document_highlight", { clear = true })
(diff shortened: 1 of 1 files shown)


pierrenel:
Tested the patch via AstroUpdate - it works \o/

mehalter:
This is a change in Neovim nightly currently and is not released, and therefore is *not* an officially supported version of AstroNvim. We do have a branch that is working on Neovim 0.8 support for when it is released and we can tag AstroNvim v2.0.0 (#391)

These changes are not supported and I cannot guarantee stability, testing, or how fast changes get adopted with each daily released nightly's breaking changes. For those reasons, I would highly highly recommend using a stable release of Neovim.

Fire-Dragon-DoL:
Notice that `user_example` template still uses `client.resolved_capabilities`

PR #788, opened 2022-07-25
fix: fix overlapping for deleteon and dashboard

blakmer:


mehalter:
@blakmer is there a reason we need to add `D` as a global delete binding and not as a personal preference in a user configuration when `d` is a default delete?

mehalter:
I'm going to go ahead and close this. I don't think it is necessary since Neo-tree has default delete bindings.

PR #789, opened 2022-07-25
mappings and which-key

### LINKED ISSUE #781: Align which-key and mappings.lua configuration
Hello,

I was trying to add some mappings to `mappings.lua` in this format:

```lua
		["<leader>"] = {
			x = {
				e = {
					function()
						require("dapui").eval()
					end,
					desc = "Evaluate Line",
				},
			},
		},
```

After much digging, It seems like this is supported in only `which-key` registration.

Keep the same syntax support for both `mappings.lua` and `which-key` or only support one style. Two is just confusing and probably more maintenance. 

**Additional context**

osamuaoki:
This addresses issues discussed at #781

Resolves #781

### DIFF
--- a/lua/user_example/init.lua
+++ b/lua/user_example/init.lua
@@ -140,21 +140,6 @@ local config = {
     },
   },
 
-  -- Modify which-key registration
-  ["which-key"] = {
-    -- Add bindings
-    register_mappings = {
-      -- first key is the mode, n == normal mode
-      n = {
-        -- second key is the prefix, <leader> prefixes
-        ["<leader>"] = {
-          -- which-key registration table for normal mode, leader prefix
-          -- ["N"] = { "<cmd>tabnew<cr>", "New Buffer" },
-        },
-      },
-    },
-  },
-
   -- CMP Source Priorities
   -- modify here the priorities of default cmp sources
   -- higher value == higher priority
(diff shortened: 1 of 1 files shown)


lua/user_example/init.lua
```diff
@@ -216,16 +201,59 @@ local config = {
...
+      ["<leader>bd"] = { "<cmd>BufferLineSortByDirectory<cr>", desc = "Sort by directory" },
+      ["<leader>bD"] = { "<cmd>BufferLineSortByRelativeDirectory<cr>", desc = "Sort by rel. dir." },
+      ["<leader>be"] = { "<cmd>BufferLineSortByExtension<cr>", desc = "Sort by extention" },
+      ["<leader>bf"] = { "<cmd>Telescope buffers<cr>", desc = "Find" },
+      ["<leader>bj"] = { "<cmd>BufferLinePick<cr>", desc = "Pick to jump" },
+      ["<leader>bl"] = { "<cmd>BufferLineCloseLeft<cr>", desc = "Close left-side" },
+      ["<leader>bn"] = { "<cmd>BufferLineCycleNext<cr>", desc = "Next" },
+      ["<leader>bp"] = { "<cmd>BufferLineCyclePrev<cr>", desc = "Previous" },
+      ["<leader>br"] = { "<cmd>BufferLineCloseRight<cr>", desc = "Close right-side" },
+      ["<leader>bt"] = { "<cmd>BufferLineSortByTabs<cr>", desc = "Sort by tabs" },
```
mehalter, line 226: can We remove these and only leave a few examples in the user_example commented out.

lua/user_example/init.lua
```diff
@@ -216,16 +201,59 @@ local config = {
...
     n = {
       -- second key is the lefthand side of the map
-      ["<C-s>"] = { ":w!<cr>", desc = "Save File" },
+      -- Terminal
+      -- Install ncdu package: (Disk size w/o snapshot btrfs subvolumes and .git)
+      ["<leader>tu"] = { function() astronvim.toggle_term_cmd "ncdu -x --exclude .git" end, desc = "Term for NCDU w/o junks" },
+      -- Install mc package:
+      ["<leader>tm"] = { function() astronvim.toggle_term_cmd "mc . ~" end, desc = "Term for MC" },
+      -- Install libreply-perl + libterm-readline-gnu-perl packages:
+      ["<leader>tp"] = { function() astronvim.toggle_term_cmd "reply" end, desc = "Term for perl5 REPL" },
```
mehalter, line 213: could we remove thse and leave the `tu` example for a lua function example?

mehalter (changes requested):
Thanks for the contribution! This is a great addition to the `user_example` documentation. I have included some comments to remove some of these examples that don't add anything except repetition. I really like the additional comments and explanations :) Just want to keep the provided `user_example` as relatively minimal.

lua/user_example/init.lua
osamuaoki: Yes

lua/user_example/init.lua
osamuaoki: Yes

osamuaoki:
I think Buffer example with less binding is needed here.  What do you think updated one.

mehalter (changes requested):
Looks great! Could you please squash these commits to a single commit and reword the commit message to fit the conventional commit standard? Probably something like `chore(user_example): make mappings example more clear`

mehalter:
@osamuaoki have you had a chance to update the commit message on this one?

osamuaoki:
I think mappings set by which-key is OK.
But use of mappings table is not for speed but for robustness.  Its much simpler.

mehalter (approved):
LGTM

PR #813, opened 2022-08-01
Add install.sh

osamuaoki:
This is a generic installer to install AstroNvim (or any other such as
NvChad) under alternative command name and without affecting normal nvim
command execution environment using XDG_* environment variables.

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -1,0 +1,168 @@
+#!/bin/sh -e
+# command and configuration setup under XDG Base Directory Specification
+# https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+# @ NAME_ORIG       command name (original)
+# @ NAME_ALT        command name (alternative under XDG)
+# @ REPO_NAME       config var for git repo name (w/ or w/o "https://github.com/")
+# @ CLONE_NAME      config var for git repo checkout dir name
+cd >/dev/null || true
+HOME=$(pwd)
+help () {
+  echo "syntax: ${0##*/} NAME_ALT"
+  echo "syntax: ${0##*/} NAME_ALT REPO_NAME"
+  echo "syntax: ${0##*/} NAME_ORIG NAME_ALT REPO_NAME"
+  echo "syntax: ${0##*/} NAME_ORIG NAME_ALT REPO_NAME CLONE_NAME"
+  echo
+  echo "Set up a NAME_ALT command executed under the XDG based alternative"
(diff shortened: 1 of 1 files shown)


osamuaoki:
Hi,

This is what I use to set up `avim` for AstroNvim and `bvim` for NvChad.  So I can break normal `~.config/nvim` contents without worry.  I intentionally skipped normal install to keep this script simple.

If you think it is not worth including, please close this feature addition PR.

If you think README.md is required for this directory `installer`, let me know.

windowsrefund:
This is a really poorly written shell script. My advice would be to get some feedback first regarding the exact problem you are trying to solve before writing a shell script that people would then be expected to run on their systems.

osamuaoki:
Let me present a use case of this script for AstroNvim which can't be realized by using ~/.config/astronvim/lua/user.

Use of this script enables you to install and test a newer upstream commit of AstroNvim without touching the current working installation.  This is quite useful to keep up development branch while always have a working installation. 

This is why comment made by windowsrefund in this thread and at https://github.com/AstroNvim/astronvim.github.io/issues/12 in which a value judgement is placed as "poor" isn't backed by sufficient facts.  Of course, there may be some "poor" points in my code but please be specific.
 
Actually, I think question is do we need to include this to be installed for everyone?  Whoever can use this may want to change details for their needs.  So I think this may be a type of script best offered as an *example script*.

installer/install.sh
```diff
@@ -0,0 +1,168 @@
...
+    REPO_NAME="https://github.com/$REPO_NAME"
+  fi
+  if [ "$REPO_NAME" = "${REPO_NAME%.git}" ]; then
+    REPO_NAME="$REPO_NAME.git"
+  fi
+fi
+
+# initialize
+BIN_DIR="$HOME/bin"  # may be ~/.local/bin
+XDG_CONFIG_HOME="$HOME/.config/$NAME_ALT"
```
JCallicoat, line 83: These should use parameter expansion to respect the `XDG_*` base directories if they have been set by the user to something other than the defaults., e.g.,

```suggestion
XDG_CONFIG_HOME="${XDG_CONFIG_HOME:-$HOME/.config}/$NAME_ALT"
```

osamuaoki:
You are right

Although this will make this script more complicated I agree this is probably a right way

Since some people already complained my script to be unnecessary complicated, let me if we get similar response from others

I would also like to hear your opinion on the approach presented by me.

I am away from PC to access this repo
Give me 2 weeks before acting on this

Sent from iPhone

PR #817, opened 2022-08-02
chore(user_example): make mappings example more clear

osamuaoki:
This should address use of mappings and ["which-key"].

This addresses issues discussed at #781

### DIFF
--- a/lua/user_example/init.lua
+++ b/lua/user_example/init.lua
@@ -153,21 +153,6 @@ local config = {
     },
   },
 
-  -- Modify which-key registration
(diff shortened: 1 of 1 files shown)


osamuaoki:
Hi,

This is the same update as the one you applied to main but this is for v2.
5dfeb77 ("chore(user_example): make mappings example more clear", 2022-07-25)

mehalter:
@osamuaoki this doesn't have any breaking changes. So I'm planning on releasing this before v2.0 so it can go in main branch

PR #829, opened 2022-08-05
feat: better syntax highlighting for rust

### LINKED ISSUE #825: Rust does not have proper syntax highlighting
**Checklist**
- [x ] `:PackerSync`
- [x ] `:AstroUpdate`
- [x ] restarted AstroNvim
- [x ] Neovim version >= 0.7.0

**Describe the bug**
Rust does not have good syntax highlighting.  Many different syntax types are the exact same color which becomes very confusing.  It's the same issue for C# but I wanted to keep this issue to one language for now.
For example, functions are the same color as Types.
If/else/fn/match, etc are the same color as crates
and so on.

**To Reproduce**
1. Install the latest Neovim
2. Install the latest AstroNvim
3. Open a rust file with nvim

**Expected behavior**
I expect clear distinction of colors between syntax types, just like you would find in VSCode or CLion

**Screenshots**
image
image

**Additional context**
Screenshots are taken from rustlings's src/main.rs
--

kabinspace:
Fixes #825
nowwww

### DIFF
--- a/lua/default_theme/treesitter.lua
+++ b/lua/default_theme/treesitter.lua
@@ -54,6 +54,11 @@ local treesitter = {
   markdownTSLiteral = { fg = C.green },
   markdownTSPunctSpecial = { fg = C.red },
   markdownTSPunctDelimiter = { fg = C.fg },
+  rustTSType = { fg = C.yellow },
+  rustTSFuncMacro = { fg = C.orange },
+  rustTSNamespace = { fg = C.yellow },
+  rustTSConstant = { fg = C.cyan },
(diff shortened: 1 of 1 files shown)


mehalter (approved):
LGTM

RockyGitHub:
Thank you very much!! You guys are amazing and I am so thrilled with what you have created

RockyGitHub:
> 

Fantastic!  I'm just so familiar with those colors at this point that it is hard to transition.  AstroNvim just gets better and better :)

PR #860, opened 2022-08-09
Carbon Language: supported added

thamognya:
- added to offical vim plugin for Carbon lang

Signed-off-by: Thamognya Kodi <contact@thamognya.com>


mehalter:
I think we would like to keep language specific plugins limited to user configurations.

PR #873, opened 2022-08-11
Add missing range code action mapping

lukaskwkw:
Hey when I was switching from VSCode to Astro I noticed that I couldn't refactor code when being in visual mode. I used that feautre very often in VSC when was working with Rust code

### DIFF
--- a/lua/configs/lsp/handlers.lua
+++ b/lua/configs/lsp/handlers.lua
@@ -34,6 +34,9 @@ astronvim.lsp.on_attach = function(client, bufnr)
         ["]d"] = { function() vim.diagnostic.goto_next() end, desc = "Next diagnostic" },
         ["gl"] = { function() vim.diagnostic.open_float() end, desc = "Hover diagnostics" },
       },
+      v = {
+        ["<leader>la"] = { function() vim.lsp.buf.range_code_action() end, desc = "Range LSP code action" },
+      },
     }),
     { buffer = bufnr }
   )


lua/configs/lsp/handlers.lua
```diff
@@ -34,6 +34,10 @@ astronvim.lsp.on_attach = function(client, bufnr)
         ["]d"] = { function() vim.diagnostic.goto_next() end, desc = "Next diagnostic" },
         ["gl"] = { function() vim.diagnostic.open_float() end, desc = "Hover diagnostics" },
       },
+      v = {
+        ["<leader>la"] = { function() vim.lsp.buf.range_code_action() end, desc = "Range LSP code action" },
+        ["<leader>lf"] = { function() vim.lsp.buf.range_formatting() end, desc = "Range Format code" },
```
mehalter, line 39: Could you uncapitalize the word "format" here

mehalter (changes requested):
Thanks for your contribution. Could you please edit your commit to have the change I requested? Also could you please reword your commit message to follow Conventional Commit. If you have questions about how to do this, there are instructions in the Contribution Guidelines. 

The commit message should be something like this: `feat(lsp): add visual mode lsp mappings`

lua/configs/lsp/handlers.lua
lukaskwkw: Sure thing! Just fixed - hope it's all good now

mehalter:
@lukaskwkw have you tested the format range function? I don't think it is working correctly.

lukaskwkw:
@mehalter 
You right. Just tested it and unfortunately it's not working. I removed from change then.

mehalter:
@lukaskwkw I just opened up #874 which fixes the LSP code formatting binding as well as adds a which-key menu label for the visual mode bindings

PR #890, opened 2022-08-14
fix(docs): update neovim badge to 0.8 in README.md

jayp0521:


mehalter (changes requested):
Could you reword the commit to say "chore" instead of fix? This isn't really a bug

PR #940, opened 2022-09-04
fix(colorizer): fix Colorizer filetype error when opening neovim

Ryymm:
After the last colorizer update, the following error started appearing when opening neovim:
`colorizer: Invalid type for filetypes "*"`

Fix making a modification to the `colorizer.lua` file


mehalter (changes requested):
This look great! Thanks for the fix with the recent changes to the API. Could you please reword the commit message to follow Conventional Commit specification? Instructions to do this are in the Contributing docs :)

Ryymm:
I have reworded the commit

mehalter (approved):
LGTM

PR #976, opened 2022-09-17
fix: auto-quit autocmd behavior

### LINKED ISSUE #975: auto_quit autocmd produces undesired behavior
### Checklist

- [X] Using a stable version of Neovim (i.e. not neovim nightly)
- [X] `:PackerSync`
- [X] `:AstroUpdate`
- [X] Restarted AstroNvim

### AstroNvim version

v1.9.2

### Neovim version (>= 0.7)

v0.7.2

### Operating system/version

Linux

### Describe the bug

I got a bug report on aerial here https://github.com/stevearc/aerial.nvim/issues/151 that I traced back to the auto_quit autocmd.
https://github.com/AstroNvim/AstroNvim/blob/9badfdd8f2e472fc79dd666679df9ed6237240a8/lua/core/autocmds.lua#L20-L27

My interpretation of this code is: "quit the current window if the total number of (visible or hidden) aerial and neo-tree buffers is equal to the number of total windows across all tabs, and that number is > 1". I don't think that makes much sense.

I was going to take a crack at fixing it, but I don't think I have a good idea of what desired behavior is. Is the intent: "close visible aerial windows in the current tab if the only other windows are neo-tree"? The original PR #869 mentions "auto close AstroNvim if a window is closed and only Aerial and Neo-tree are opened." Is that actually what you want to have happen? I don't know what the general workflow with neo-tree looks like, but as a user I feel like I might expect to be able to have a single-window file browsing experience without auto-quitting. Either way, I'm happy to make the changes if you can clarify the intent.

### Steps to Reproduce

1. `nvim .`
2. open a file
3. `:AerialToggle`

See error; aerial is not open.

### Expected behavior

No error

### Screenshots

_No response_

### Additional Context

_No response_

stevearc:
fixes #975

If there is only one window, do nothing
Otherwise, if all of the visible windows are neo-tree and aerial, close the current tab (if there are other tabs) or exit vim.

### DIFF
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -17,14 +17,22 @@ cmd("BufEnter", {
   desc = "Quit AstroNvim if more than one window is open and only sidebar windows are list",
   group = "auto_quit",
   callback = function()
-    local num_wins = #vim.api.nvim_list_wins()
-    local sidebar_fts = { "aerial", "neo-tree" }
-    local sidebars = {}
-    vim.tbl_map(function(bufnr)
-      local ft = vim.api.nvim_buf_get_option(bufnr, "filetype")
-      if vim.tbl_contains(sidebar_fts, ft) then sidebars[ft] = true end
-    end, vim.api.nvim_list_bufs())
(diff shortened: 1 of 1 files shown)


mehalter (approved):
Thanks so much for this contribution!

PR #988, opened 2022-09-20
fix(default_theme): better bufferline integration and external themes…

shans10:
Updated default_theme with better bufferline integration. This is how it looks now
Screenshot from 2022-09-20 19-06-13


mehalter:
@shans10 the reason we have been not been using these setup functions is because it breaks the ability to change themes and the bufferline look good. This works if you just use one theme on startup and that's it, but if you change the theme while using astronvim then the colors do not switch. I think for now we are going to stick with just using the highlight group approach.

PR #1000, opened 2022-09-24
fix(laststatus): respect user configurations

pedromss:
Setting `vim.opt.laststatus` in init.vim through the `g` or `opt` keys isn't respected when we leave the Alpha page

### DIFF
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -58,7 +58,8 @@ if is_available "alpha-nvim" then
     group = "alpha_settings",
     pattern = "alpha",
     callback = function()
-      local prev_status = vim.opt.laststatus
+      local user_opts = require "user.options"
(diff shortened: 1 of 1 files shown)


mehalter:
@pedromss I don't think I can replicate this. All the `user.options` table does is set `vim.` variables. So this is no different. I will say that vim.g.laststatus shouldn't be an actual setting. Make sure you are doing `user.options.opt.laststatus` which would be `vim.opt.laststatus`. Let me know if that fixes your incorrect behavior. If not I can investigate some more to see if this PR is actually necessary.

pedromss:
@mehalter I just noticed that it was actually lualine setting this to false in the first place (in my config).

So I was setting it in `vim.opt`, then lua line was setting `laststatus` to something other than `3` and then astronvim was picking it up.

Sorry for the confusion, closing.

PR #1045, opened 2022-10-07
fix(aerial): min_width is set inside `layout`

jayp0521:
https://github.com/stevearc/aerial.nvim/blob/e0215f9dcf55e917f547ad2c688be6fe67a5926d/lua/aerial/config.lua#L13


jayp0521:
please squash and merge. no need for 2 commits

PR #1065, opened 2022-10-10
AstroNvim Nightly

mehalter:


mehalter:
This has some pretty important bug fixes and needs to be merged and released. Because there is a new feature for toggling the tabline this will release to v2.1.0

PR #1080, opened 2022-10-11
Revert the default behavior for `<C-(p|n|k|j)>` mapping in `cmp` to `Insert`

yehorb:
Hi.

Looks like the default behavior for `<C-(p|n|k|j)>` mapping in `cmp` was changed to `Select` in v2.x, but it was `Insert` in v1.10.0. Technically, the function call to `cmp.mapping.select_(prev|next)_item()` had no arguments in v1.10.0, but it defaults to `Insert` in the `cmp` source code.

I was very confused when my completion stopped working as expected after v2.x upgrade. I consider this behavior a breaking change, but it was not mentioned in the migration guide.

This PR changes the default `AstroNvim` mappings to be similar to [`cmp's insert preset`](https://github.com/hrsh7th/nvim-cmp/blob/53bd5749011c0830d54d9b22a37259651d5916df/lua/cmp/config/mapping.lua#L36), where `<Up> and <Down>` use `Select` behavior and `<C-(p|n)>` use `Insert` behavior. Additional `<C-(j|k)>` mapping use the `Insert` behavior.

It is still a change from v1.10.0, but a lesser one, and I think behaviors being clearly defined in `AstroNvim` source is an upgrade overall, so not going back to no argument calls.

Also, it looks like there was a copy-paste mistake, as `cmp.mapping.select_(prev|next)_item()` only expects a single argument, and the `{ "i", "c" }` argument has no significance in these calls.


yehorb:
A small reminder. `cmp.SelectBehavior.Select` means that selecting the completion item in the popup menu does not change the buffer, you still need to press `<CR>` to trigger completion. `cmp.SelectBehavior.Insert` changes the buffer on the selection of the item, you do not need to press `<CR>` to complete.

mehalter:
I like this, thank you for the contribution @yehorb ! Since both of these commits are effectively contributing to the same thing, could you please squash your commits. Also when you squash them could you please reword the final commit to include the part of the code you are modifying? This would look something like `chore(cmp): revert default completion selection behavior in insert mode` (Also note that I think this is more of a chore or refactor rather than a fix since there is no bug here)

yehorb:
@mehalter, I rebased the changes into the single commit.

mehalter:
Yes please @yehorb we have a stylua config file in the repo so if you have stylua you can run it to fix the issues.

mehalter:
@yehorb thanks for that. Can you please rebase your commits back to a single commit and keep the text of the first commit describing the refactor. We do not need separate commits for formatting.

yehorb:
@mehalter, I applied `stylua` formatting for my changes (and added them into original commit), but it also fired on unrelated files (I added separate `chore` commit for that).

mehalter:
oh I see I see, interesting

mehalter:
Sorry about that thank you!

PR #1121, opened 2022-10-17
feat(utils): add pcall error message when loading module fails

matharman:
Hi!

When trying to debug a problem loading my `user/init.lua` file, I found that without this error message I had insufficient information to chase down what I'd written wrong in the config.

I know some `pcall(require, ...)` errors can be pretty long, so if this error message was left out on purpose, I'd be happy to help figure out an alternative that still provides more error info.


mehalter (approved):
This looks great, thanks so much!

PR #1139, opened 2022-10-20
fix(lsp): fix format_on_save whitelist

maxbrunet:
I do not know what happened, but the commit that ended up in Nightly (720db1f2487ea7f1c1081aa50ecb2dcb61cba4c7) is not the commit that was approved in #1133.

### DIFF
--- a/lua/core/utils/lsp.lua
+++ b/lua/core/utils/lsp.lua
@@ -94,7 +94,8 @@ astronvim.lsp.on_attach = function(client, bufnr)
         type(format_on_save) == "table"
         and format_on_save.enabled == true
         and (
-          next(format_on_save.allow_filetypes or {}) and vim.tbl_contains(format_on_save.allow_filetypes, filetype)
+          next(format_on_save.allow_filetypes or {}) ~= nil
+            and vim.tbl_contains(format_on_save.allow_filetypes, filetype)
           or not vim.tbl_contains(format_on_save.ignore_filetypes or {}, filetype)
         )
       )


mehalter:
@maxbrunet The recommended fix actually doesn't work either. Just for reference, if something is `== nil` then it is resolved as false, so you don't necessarily need the `nil` check here as it is implied. I have tested this branch and it doesn't resolve the issues. I have rewritten this to be a bit cleaner and have resolve the issue in the `nightly` branch with commit https://github.com/AstroNvim/AstroNvim/commit/dff7fc2f3388ded10f66dac3b6644e3c85824194

Let me know what you think and I will go ahead and tag this patch release to fix it :)

maxbrunet:
> I have tested this branch and it doesn't resolve the issues.

I do not know how I do my testing, I am very confused by the results I get. :see_no_evil:

> I have rewritten this to be a bit cleaner and have resolve the issue in the `nightly` branch with commit [dff7fc2](https://github.com/AstroNvim/AstroNvim/commit/dff7fc2f3388ded10f66dac3b6644e3c85824194)

Thank you, I am indeed not well versed in Lua or even Vim config, I looked for soemthing like `vim.tbl_empty()` and couldn't find it. I tested it and it looks good (I believe :laughing:)

mehalter:
Yeah I haven't found a good way really to find built in lua api commands that Neovim contains. The best I have found is this page: https://neovim.io/doc/user/lua.html and I just ctrl-f for words I think I should be able to find like "empty" :joy: there are a fair number of `tbl_` commands that are quite useful :) Thanks for reporting the issue tho! Gave me a good excuse to really clean up the auto formatting code as well haha

mehalter:
If you are curious, the reason was because of the `__ and __ or __` notation instead of `if/else`

in an `if/else` scenario:

```lua
if __ then
  return __
else
  return __
end
```

if the if condition is satisfied then you return the result of that `if` block and never touch the `else` block, but that's not how the boolean logic works for `__ and __ or __`

so:

```lua
if true then
  return false
else
  return true
end
```

this will return true, but `true and false or true` will return true because it evaluates `true and false` first and then it parses `false or true` which will be `true`

Not sure if that's a clear explanation :joy: sorry if it just made everything more confusing

maxbrunet:
Yeah, I get the basic boolean algebra

From my `reflog`, the weird thing is this is how I did the first time around:

```diff
diff --git a/lua/core/utils/lsp.lua b/lua/core/utils/lsp.lua
index 74d226f..cc13ba8 100644
--- a/lua/core/utils/lsp.lua
+++ b/lua/core/utils/lsp.lua
@@ -87,12 +87,14 @@ astronvim.lsp.on_attach = function(client, bufnr)
       { desc = "Format file with LSP" }
     )
     local format_on_save = astronvim.lsp.formatting.format_on_save
+    local filetype = vim.api.nvim_buf_get_option(bufnr, "filetype")
     if
       format_on_save == true
       or (
         type(format_on_save) == "table"
         and format_on_save.enabled == true
-        and not vim.tbl_contains(format_on_save.ignore_filetypes or {}, vim.api.nvim_buf_get_option(bufnr, "filetype"))
+        and vim.tbl_contains(format_on_save.allow_filetypes or { filetype }, filetype)
+        and not vim.tbl_contains(format_on_save.ignore_filetypes or {}, filetype)
       )
     then
       local autocmd_group = "auto_format_" .. bufnr
```

But my testing... I got confused looking for that "is empty" function, maybe that `or { filetype }` was too lazy :shrug:, then my brain stopped working

maxbrunet:
I think my poor testing comes from the Lua cache ([lewis6991/impatient.nvim](https://github.com/lewis6991/impatient.nvim)) which is not always invalidated properly when files are managed by Nix.

I'm going to add an `onChange` hook to my `home-manager` config clear the cache automatically: 
```shell
nvim --headless +LuaCacheClear +quitall
```

PR #1143, opened 2022-10-20
fix(ui): show changed status for toggling autoformatting and tabline

dohsimpson:
Before: 
image

After:
image


mehalter (approved):
Ah thanks so much for catching this! LGTM

PR #1159, opened 2022-10-22
code-action-menu

Aprilswind:
Integration of code-action-menu


mehalter:
This plugin is still considered beta and unstable so I don't think that it is a good fit for AstroNvim at the moment. Thanks for the suggestion though, I'll keep my eye out on it.

PR #1167, opened 2022-10-23
Custom

meyerhot95:


mehalter:
I'm not sure what this is. Maybe to keep track of the difference between your config and the base. I would definitely recommend not forking AstroNvim and making direct changes in the code if you  want to receive updates. I would recommend using the user configuration system that we have in place to maintain update-ability: https://astronvim.github.io/Configuration/manage_user_config

PR #1176, opened 2022-10-25
feat(devicons): better highlights for devicon

kabinspace:
Currently packagejson icons look ugly due to dark red colour.


mehalter (changes requested):
Nice! Could you squash these commits to a single commit, also could you change it from `feat` to `chore`. This doesn't change the API at all so it isn't a new feature really

kabinspace:
@mehalter done

mehalter (approved):
LGTM

PR #1177, opened 2022-10-25
chore:  fix typo in mappings.lua  desc

syotaro:
Lack of g in the word description

naviation→navigation


mehalter (changes requested):
Thanks for this edit! Could you please remove unicode characters from the commit message and also use the keyword `chore` for this type of edit? This just makes sure that the content of the commit is compatible with all terminals and fonts.

syotaro:
@mehalter 
We have addressed the points raised.
(I also got to know why. That's a learning experience.)

mehalter (approved):
LGTM

PR #1193, opened 2022-10-27
investigating nvim-dap integration

mehalter:
Not sure how fast this will be finished. It needs to  go through a LOT a LOT of testing and a lot of documentation writing before we can get this merged, but with some new developments to [`mason-nvim-dap.nvim`)[https://github.com/jayp0521/mason-nvim-dap.nvim] thanks to the hard work from @jayp0521 we are much much closer to providing a user friendly way to integration debugging into AstroNvim!

Here is a small preview so far using a fresh  installation of AstroNvim installing treesitter, language servers, linters, formatters, and debuggers (new) and being able to run the debugger into files without having to even restart the editor: 
https://asciinema.org/a/WQfjL3JrkNINTThergxnKZzoK

### TODO

- [x] Finalize Mappings
- [x] Add highlight groups for `nvim-dap` and `nvim-dap-ui` to `default_theme`


lua/configs/which-key-register.lua
```diff
@@ -36,6 +37,8 @@ if is_available "telescope.nvim" then
   init_table("n", "<leader>", "g")
 end
 
+if is_available "nvim-dap" then init_table("n", "<leader>", "D") end
```
jayp0521, line 40: capitals are annoying. anyway this could be change? I know the dashboard mapping would be backwards incompatible, possibly

lua/core/mappings.lua
```diff
@@ -193,6 +193,58 @@ if is_available "toggleterm.nvim" then
...
+      require("dap").terminate()
+    end,
+    desc = "Debugger: Stop",
+  }
+  maps.n["<F29>"] = { function() require("dap").restart_frame() end, desc = "Debugger: Restart" }
+  maps.n["<F6>"] = { function() require("dap").pause() end, desc = "Debugger: Pause" }
+  maps.n["<F9>"] = { function() require("dap").toggle_breakpoint() end, desc = "Debugger: Toggle Breakpoint" }
+  maps.n["<F10>"] = { function() require("dap").step_over() end, desc = "Debugger: Step Over" }
+  maps.n["<F11>"] = { function() require("dap").step_into() end, desc = "Debugger: Step Into" }
+  maps.n["<F23>"] = { function() require("dap").step_out() end, desc = "Debugger: Step Out" }
```
jayp0521, line 218: do these originate from somewhere? `F29` seems atypical

lua/core/mappings.lua
```diff
@@ -193,6 +193,58 @@ if is_available "toggleterm.nvim" then
   maps.t["<C-'>"] = maps.n["<F7>"]
 end
 
+if is_available "nvim-dap" then
```
jayp0521, line 196: it would be interesting to see if `mappings.lua` could more closer to the config for each plugin. way out of the scope of this PR ofc

lua/core/mappings.lua
```diff
@@ -193,6 +193,58 @@ if is_available "toggleterm.nvim" then
...
+    desc = "Debugger: Stop",
+  }
+  maps.n["<F29>"] = { function() require("dap").restart_frame() end, desc = "Debugger: Restart" }
+  maps.n["<F6>"] = { function() require("dap").pause() end, desc = "Debugger: Pause" }
+  maps.n["<F9>"] = { function() require("dap").toggle_breakpoint() end, desc = "Debugger: Toggle Breakpoint" }
+  maps.n["<F10>"] = { function() require("dap").step_over() end, desc = "Debugger: Step Over" }
+  maps.n["<F11>"] = { function() require("dap").step_into() end, desc = "Debugger: Step Into" }
+  maps.n["<F23>"] = { function() require("dap").step_out() end, desc = "Debugger: Step Out" }
+  maps.n["<leader>Db"] = { function() require("dap").toggle_breakpoint() end, desc = "Toggle Breakpoint (F9)" }
+  maps.n["<leader>DB"] = { function() require("dap").clear_breakpoint() end, desc = "Clear Breakpoints" }
```
LinuCC, line 220:
```suggestion
  maps.n["<leader>DB"] = { function() require("dap").clear_breakpoints() end, desc = "Clear Breakpoints" }
```

On my end only `clear_breakpoints` exist

lua/core/mappings.lua
mehalter: Please see the comment at the top of this comment block, these atypical function numbers are how you code in modifiers onto function keys and is how terminals and neovim translate these key codes. For example `<F29>` is how you would say `<C-F5>`. Using just `<C-F5>` does not work in this case.

lua/configs/which-key-register.lua
mehalter: I agree, but in this case I think it will be important to not impose breaking changes. This also aligns with how VS Code configures debugging bindings which is a nice unification for people migrating back and forth. The main thing here is also for most used actions such as toggling break points/starting/stopping are all accessible through function keys which are much more convenient but take learning. For this reason we have added hints to the function key equivalent to relevant mappings' descriptions.

Outside of this I'm not sure if there is a good mapping other than `D` that would be applicable here. If you have a recommendation I would love to hear! (without breaking change of mapping `d`. We could also push this to AstroNvim v3 which I'm totally fine with and do the breaking change of `<leader>d`

A-Lamia:
I think maybe using [Hydra](https://github.com/anuvyklack/hydra.nvim) instead of witch-key, might be better, Hydra is able to have modes be triggered so you could potentially trigger a Debug mode, with out constantly pressing the leader key. This would also allow having a settings menu that persists. 

Just some thoughts.

jayp0521:
> I think maybe using [Hydra](https://github.com/anuvyklack/hydra.nvim) instead of witch-key, might be better, Hydra is able to have modes be triggered so you could potentially trigger a Debug mode, with out constantly pressing the leader key. This would also allow having a settings menu that persists. 
> 
> Just some thoughts.

This would be interesting, but I doubt astronvim should vend this. Maybe make it easy to use though

jayp0521:
unsure how you feel about adding this: https://github.com/theHamsta/nvim-dap-virtual-text. I like it, but not a necessity

mehalter:
I've thought about this as well as an out of the box functionality. I definitely like it, but I'm kinda leaning towards providing just dap and the UI and letting other users decide if they want the virtual text. Some people don't like the virtual text even for diagnostics because it clutters the editor. I'm thinking to leave it out especially since the dap-ui does provide a way to easily to see variable values already

jayp0521:
I have used this for a bit. Seems stable to me. Anything else blocking this from merging?

mehalter:
I need to take some time and finalize everything a bit. Sorry @jayp0521 I haven't had time to test everything recently. Also with the new lazy loading, I wanted to make sure that it was running well and getting all the kinks there ironed out first stability wise. Also I want to write a decent amount of docs regarding this since most users that want to use this will need to do some configuration at least a bit until more language servers are added. This also should maintain a list of LSP specific plugins that provide DAP configurations and make sure they play nicely with the mason stuff like how we have the LSP stuff working.

Sorry this is a lot of information, also using it as a brain dump of ideas into text to refer to later 😂

jayp0521:
> I need to take some time and finalize everything a bit. Sorry @jayp0521 I haven't had time to test everything recently. Also with the new lazy loading, I wanted to make sure that it was running well and getting all the kinks there ironed out first stability wise. Also I want to write a decent amount of docs regarding this since most users that want to use this will need to do some configuration at least a bit until more language servers are added. This also should maintain a list of LSP specific plugins that provide DAP configurations and make sure they play nicely with the mason stuff like how we have the LSP stuff working.
> 
> Sorry this is a lot of information, also using it as a brain dump of ideas into text to refer to later 😂

What are your thoughts on merging this in with everything disabled by default? It would decrease your effort in maintaining a separate branch and people might be inclined to opt into this feature

mehalter:
Ah I think I'm a fan of this, I'll do this tomorrow @jayp0521 and get it merged to `nightly` at least with disabled by default

mehalter:
I will spend some time tomorrow finalize the mappings and stuff and get something merged tomorrow hopefully. If not, Early next week

jayp0521:
Awesome, thanks!

mehalter:
I'm going to go ahead and move this to nightly and just work on it from there

PR #1198, opened 2022-10-28
chore(default_theme): Add missing Aerial icon

hunger:
This stops areial from erroring out three times on each cursor move, asking for AerialTypeIcon :-)


mehalter (changes requested):
Thanks for catching this! Could you please reword the commit to say `chore(default_theme}` rather than just `chore`? :)

mehalter (approved):

PR #1199, opened 2022-10-28
chore(mapping): Fix typo in DAP keymap

hunger:


mehalter (changes requested):
Thanks for this! Could you please reword the commit to be `chore(mappings)` rather than just `chore`?

mehalter (approved):

PR #1202, opened 2022-10-29
fix: unnecessary configuration option

kunzaatko:
`symbol` is already the default


mehalter (changes requested):
Thanks for this contribution!

Could you please reword your commit message to say `chore(lspkind):` instead of `fix:`. This helps to keep the git log easy to understand where things are touching the code. Also this is a fix because it doesn't fix a bug, it's just a chore to remove unnecessary lines.

mehalter (approved):
LGTM

PR #1203, opened 2022-10-29
autopairs

kunzaatko:
- fix: `TelescopePrompt` is disabled by default
- fix: disable `on_confirm_done` for `tex`


mehalter (changes requested):
Thanks for this contribution! Last thing is could you please squash these commits to a single commit with the message similar to `chore(autopairs): refactor some configuration options`

mehalter (approved):
LGTM

PR #1233, opened 2022-11-03
fix(neo-tree): add Mapping `h` to toggle_hidden

ypli0629:
the default `toggle_hidden` mapping was overrided by `prev_source`


mehalter (approved):
Thanks so much for catching this! So sorry it slipped my mind! LGTM

PR #1261, opened 2022-11-06
ENH - Add `spaceless` to remove trailing whitespace

PABannier:
This PR adds the `spaceless` package to the plugins to remove trailing whitespace.


mehalter:
Many formatters already do this with LSP and Null-ls. So I don't think this is necessary to upstream. Feel free to put it in your user configuration though!

PR #1277, opened 2022-11-08
Nightly

mehalter:


lua/core/utils/init.lua
```diff
@@ -287,8 +287,8 @@ function astronvim.toggle_term_cmd(term_details)
   local term_key = term_details.cmd
   -- set the count in the term details
   if vim.v.count > 0 and term_details.count == nil then
-    term_details.count = vim.v.count
-    term_key = term_key .. vim.v.count
+    term_details.count = vim.v.count + 1000
```
vtvz, line 290: I'm not sure this will help as `vim.v.count > 0` almost always is not the case. I think `if` condition should be removed entirely

mehalter: Resolved this in latest revision of the commit

kabinspace (approved):
LGTM

PR #1285, opened 2022-11-10
feat : removing trailing spaces

chanchann:
I think it is good to add a function to remove the trailing white spaces in files.


mehalter:
Thanks for the recommendation! This is handled by most LSP formatters already. I'm not sure if we need to add an autocommand to do it to every file. You can also place this in your `polish` function in your `user/init.lua` file to put it in your own user configuration

chanchann:
Ok, thanks.

PR #1305, opened 2022-11-13
Nightly

mehalter:


A-Lamia:
Works perfectly!

mehalter:
Awesome thanks for testing and reporting that bug! That would have definitely slipped through the cracks @A-Lamia

A-Lamia:
@mehalter Just doing my part to keep awesome software running, thanks for the hard work! ;)

PR #1318, opened 2022-11-15
Note about cleanup previous installation

kazhuravlev:


.github/README.md
```diff
@@ -79,6 +79,12 @@ AstroNvim is an aesthetic and feature-rich neovim config that is extensible and
 mv ~/.config/nvim ~/.config/nvimbackup
 ```
 
+### Remove cache from the previous installation
+
+```
+rm -rf ~/.local/share/nvim
```
mehalter, line 85: Could you please change this to `mv ~/.local/share/nvim ~/.local/share/nvim.bak` just so we aren't having users blindly deleting directories if they aren't necessarily paying attention?

mehalter (changes requested):
Thanks for this contribution! Could you please check out the one comment I made and also can you please reword your commit message to match Conventional Commit as described in the contribution guide (https://github.com/AstroNvim/AstroNvim/blob/main/.github/CONTRIBUTING.md)

It should be something like `chore(readme): improve installation guide`

.github/README.md
kazhuravlev: done

kazhuravlev:
@mehalter Is it better now?

mehalter (approved):
LGTM! Thanks for the contribution!

PR #1354, opened 2022-11-19
fix(cmp): disable lspkind symbols with text

kabinspace:


mehalter:
This is already fixed in nightly

PR #1382, opened 2022-11-26
fix(treesitter): correct float color

kabinspace:
Float number color needs to be the same as of integer number.


mehalter (approved):
LGTM

PR #1389, opened 2022-11-29
feat!: support for multiple user based configurations

jayp0521:
removes astronvim.user_settings global variable. I looked for a way around this, but did not see a nice way. Open to suggestions

### DIFF
--- a/lua/core/utils/init.lua
+++ b/lua/core/utils/init.lua
@@ -49,8 +49,6 @@ local function load_module_file(module)
   return found_module
 end
 
---- user settings from the base `user/init.lua` file
-astronvim.user_settings = load_module_file "user.init"
 --- default packer compilation location to be used in bootstrapping and packer setup call
 astronvim.default_compile_path = stdpath "data" .. "/packer_compiled.lua"
 --- table of user created terminals
@@ -188,10 +186,14 @@ end
 
 --- Search the user settings (user/init.lua table) for a table with a module like path string
 -- @param module the module path like string to look up in the user settings table
+-- @param prefix a module prefix for where to search (default: "user")
 -- @return the value of the table entry if exists or nil
-local function user_setting_table(module)
+local function user_setting_table(module, prefix)
+  prefix = prefix or "user"
+
   -- get the user settings table
-  local settings = astronvim.user_settings or {}
+  local settings = load_module_file(prefix .. ".init") or {}
+
   -- iterate over the path string split by '.' to look up the table value
   for tbl in string.gmatch(module, "([^%.]+)") do
     settings = settings[tbl]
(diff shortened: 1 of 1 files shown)


jayp0521:
https://github.com/AstroNvim/AstroNvim/issues/1388

lua/core/utils/init.lua
```diff
@@ -49,8 +49,6 @@ local function load_module_file(module)
   return found_module
 end
 
---- user settings from the base `user/init.lua` file
-astronvim.user_settings = load_module_file "user.init"
```
jayp0521, line 53: unsure how to keep this

jayp0521: note: I searched all of github and did not see any non-fork using this variable

lua/core/utils/init.lua
```diff
@@ -270,20 +272,29 @@ end
...
-  -- try to load a module file if it exists
-  local user_settings = load_module_file((prefix or "user") .. "." .. module)
-  -- if no user module file is found, try to load an override from the user settings table from user/init.lua
-  if user_settings == nil and prefix == nil then user_settings = user_setting_table(module) end
-  -- if a user override was found call the configuration engine
-  if user_settings ~= nil then default = func_or_extend(user_settings, default, extend) end
-  -- return the final configuration table with any overrides applied
+
+  -- if nil, then default to { "user" }
+  prefixes = prefixes or { "user" }
```
jayp0521, line 284: oh right, there needs to be a way to actually override this easily. in my testing, I just override it directly in the file, which obviously is ready for merge. any suggestions?

jayp0521: @mehalter

mehalter:
Hm, I will have to think about this. What is the workflow here for switching the user folder that is being used? My biggest concern is definitely the usability/enhancement it provides as well as the breaking change. As it stands now this PR would require bumping to AstroNvim v3...

Also I'm wondering if this is a space we want to move into the scope of AstroNvim. My initial thoughts are this can come down to the user's dotfiles management. There are a lot of good dotfile managers that are able to do multi Machine configurations with differences.

lua/core/utils/init.lua
mehalter: Hm, yeah keeping this would definitely be more ideal so we can avoid tagging v3 of AstroNvim so soon

lua/core/utils/init.lua
mehalter: Yeah this needs to be configurable without modifying the core files. This is completely unsupported with our updater and is not going to be supported in the future. We need to have clear separation between the core files and user files

jayp0521: A shell variable is the best I can think of. not much else

mehalter:
Also now that I think about it. I think this can all be done in the current setup without any changes 👀 I will put together an example tomorrow that runs with our current setup and allows for separate configurations that all live in the current user folder. Then this could also be configured easily.

jayp0521:
> What is the workflow here for switching the user folder that is being used?

I don't think I can share my work stuff specifically, but at work, we have some internal LSPs, some special `neotest` configuration. some internal only plugins. since the internal LSPs, I primarily have in a separate file under `user/servers/`, it is not hard to maintain, but the rest gets a little confusing

> I think this can all be done in the current setup without any changes

cool, interested in how

PR #1416, opened 2022-12-02
Fix mason-null-ls' setup_handlers

Cretezy:


mehalter:
This isn't a bug, we are specifically trying to separate out the keys for the `setup` calls and outside APIs, similar to how we are doing `cmp` and `luasnip` configuration

PR #1437, opened 2022-12-12
Heirline Bufferline

mehalter:
This is just a placeholder draft PR for tracking the new bufferline rewrite with Heirline. This allows us to remove the `bufferline` plugin and use our powerful `astronvim.status` API for creating the bufferline/tabline in AstroNvim.

This will be something that won't be merged for a long while. This has some significant breaking changes and will be released with AstroNvim v3 (whenever that ends up being).


mehalter:
reworked this into a non-breaking change as an opt in feature.

PR #1466, opened 2022-12-23
Investigation of lazy.nvim

### LINKED ISSUE #1464: User Heirline for tabline by default
We would like to reduce our reliance on redundant plugins. We have an optional Heirline based bufferline currently available, we should replace the Bufferline plugin by default and just use Heirline. This gives us a lot more control over things such as colors and information displayed and gives users a lot more power over customizations.

### LINKED ISSUE #1465: Switch to Lazy Package Manager
### Is your feature related to a problem?

Packer has a lot of weirdness and things we don't like when it comes to maintenance. (1) the lazy loading in Packer is a bit complicated and makes it hard to make sure we are lazily loading plugins "correctly" and (2) the compilation phase with Packer is a bit weird and leads to a lot of issues if you don't remember to recompile the packer configuration.

### Describe the new feature

Adopt [lazy.nvim](https://github.com/folke/lazy.nvim) which has a more modern approach and a much more sane description of Lazy

### Additional context

During investigation of this it has also exposed some poor abstractions we made when it comes to configuring `cmp` and `cmp` sources. We will have to remove this as well and update the documentation for adding and configuring `cmp` sources. This in the end will also be a lot more flexible and powerful in general.

mehalter:
Resolves #1465 
Resolves #1464


mehalter:
Closing this for now

PR #1476, opened 2022-12-27
v2.11.0 Nightly

mehalter:
Add ability to disable default `vim.lsp.handlers` and improve lazy loading of `nvim-notify`

These additions make it super easy to add `noice.nvim` without having to disable any features. Here is a minimal user configuration that sets up noice:

```lua
return {
  options = {
    g = {
      lsp_handlers_enabled = false,
    },
  },
  plugins = {
    init = {
      ["folke/noice.nvim"] = { config = function() require("noice").setup() end },
    },
  },
}
```


lua/core/mappings.lua
```diff
@@ -2,8 +2,6 @@ local is_available = astronvim.is_available
 
 local maps = { i = {}, n = {}, v = {}, t = {}, [""] = {} }
```
gwww, line 3: Should `[""] = {}` be removed? I'm not even sure how this is supposed to work. I use <comma> for my leader.

mehalter: Yeah I suppose we can remove this as well since we aren't setting anything with an empty mode

PR #1499, opened 2023-01-02
fix(neo-tree): fix system_open help description

DiogoDoreto:
When pressing `?` on neo-tree, the description for `O` mapping just says `<function>`. With this fix it shows `system_open`.

Before:
image

After:
image


mehalter (approved):
Thanks! LGTM

PR #1517, opened 2023-01-09
Check for python and python3

owittek:
Enables the python terminal without requiring a python symlink if python3 is found. Not sure if changing the description is desired but I find it nice to see which if-branch is actually being used.


mehalter:
Thanks for the recommendation @owittek ! I am going to add this compatibility with a slightly different approach, but this will be included in `v3` that comes out in a couple weeks :D

PR #1530, opened 2023-01-12
fix: heirline new setup method

uggla:
- Heirline has introduced a breaking change in the setup method with commit https://github.com/rebelot/heirline.nvim/commit/7b57b27e9e4d1ffa4f3e149f8fdc927db18a850a that prevent AstroNvim from starting without errors.

This MR calls the new setup method with the correct parameters.

### DIFF
--- a/lua/configs/heirline.lua
+++ b/lua/configs/heirline.lua
@@ -282,7 +282,7 @@ local heirline_opts = astronvim.user_plugin_opts("plugins.heirline", {
       }
     or nil,
 })
-heirline.setup(heirline_opts[1], heirline_opts[2], heirline_opts[3])
+heirline.setup({statusline=heirline_opts[1], winbar=heirline_opts[2], tabline=heirline_opts[3]})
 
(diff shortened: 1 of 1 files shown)


mehalter:
This is fixed in the upcoming v3 release. I would recommend either using the stable channel of AstroNvim or if you are using nightly which is an unsupported and unreleased channel you can pick the commit manually in your user configuration

mehalter:
This is a breaking change in AstroNvim's API so this cannot be backported to a v2.X release

uggla:
> eleas

Maybe I missed something, but I have done this fix because I saw the issue on the stable branch this morning after upgrading AstroNvim + PackerSync.
PackerSync brings the new setup method from heirline code and broke Astro. I think it will also impact other users when they will upgrade.

mehalter:
You should double check what updater settings you have in your user configuration. If you are using stable then you should probably just delete them and take the defaults. Stable AstroNvim does not have this issue

PR #1573, opened 2023-02-12
fix(lsp): lua lsp is now called lua_ls

luxus:
sumneko got renamed https://github.com/neovim/nvim-lspconfig/pull/2439


mehalter (approved):
LGTM

PR #1576, opened 2023-02-12
fix(dap): not loading dap on all files

luxus:
when i open a md or a txt dap is loaded. 
this adds more then 40ms to the loadtime when a file is directly opens.
i guess dap needs some manual configuration anyways, so maybe it makes sense to let the user load it.


mehalter:
This breaks automatic installation of debuggers that you get from `mason-nvim-dap`

PR #1587, opened 2023-02-25
feat: toggle diagnostics visibility between more options

Subjective:
There is currently an option to toggle between diagnostics, but toggling "virtual text off" also removes the diagnostics from the sign column. In a lot of situations, it makes sense to turn off virtual text but continue to show line diagnostics on the side, like when the window gets too cluttered. The config files have to be updated on the user's end; instead of setting both `diagnostics_enabled` and `status_diagnostics_enabled`, the user sets a single option `diagnostics_mode`, which determines the initial visibility for diagnostics. 

This is my first ever PR and my first experience with Lua, so I would greatly appreciate any feedback or areas of improvement for future contributions. Thanks!

Screen Recording 2023-02-24 at 2 57 22 PM

### DIFF
--- a/lua/core/diagnostics.lua
+++ b/lua/core/diagnostics.lua
@@ -17,13 +17,51 @@ for _, sign in ipairs(signs) do
 end
 
 astronvim.lsp.diagnostics = {
-  off = {
+  -- diagnostics off
+  [0] = {
+    -- status = "all diagnostics off",
     underline = false,
     virtual_text = false,
     signs = false,
     update_in_insert = false,
   },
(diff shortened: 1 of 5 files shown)


mehalter (changes requested):
I like this idea! Could you please look migrating this code to modify the `v3` branch? There have been a lot of development over the last couple months since we have frozen v2 and just want to make sure we can get this PR woved over to the latest codebase! :)

Subjective:
Alright, I just migrated all the code over to v3 and opened a new PR!

PR #1588, opened 2023-02-25
fix: remove duplicated DiagnosticSignError sign

iwfan:


mehalter:
Thanks for the contribution! This has actually already been fixed in the current development branch of AstroNvim for the upcoming release of v3 next week.

PR #1590, opened 2023-02-25
feat: toggle diagnostics visibility between more options

Subjective:
There is currently an option to toggle between diagnostics, but toggling "virtual text off" also removes the diagnostics from the sign column. In a lot of situations, it makes sense to turn off virtual text but continue to show line diagnostics on the side, like when the window gets too cluttered. The config files have to be updated on the user's end; instead of setting both `diagnostics_enabled` and `status_diagnostics_enabled`, the user sets a single option `diagnostics_mode`, which determines the initial visibility for diagnostics. 

This is my first ever PR and my first experience with Lua, so I would greatly appreciate any feedback or areas of improvement for future contributions. Thanks!

Screen Recording 2023-02-24 at 2 57 22 PM


lua/astronvim/utils/lsp.lua
```diff
@@ -23,17 +23,51 @@ local setup_handlers = user_opts("lsp.setup_handlers", {
   function(server, opts) require("lspconfig")[server].setup(opts) end,
 })
 
-M.diagnostics = { off = {}, on = {} }
+M.diagnostics = { [0] = {}, [1] = {}, [2] = {}, [3] = {} }
```
mehalter, line 26: This should be simplified to just `M.diagnostics = { {}, {}, {}, {} }`

lua/astronvim/utils/lsp.lua
```diff
@@ -23,17 +23,51 @@ local setup_handlers = user_opts("lsp.setup_handlers", {
...
 
 M.setup_diagnostics = function(signs)
   M.diagnostics = {
-    off = {
+  -- diagnostics off
+    [0] = {
```
mehalter, line 31: This line should just be `{`

lua/astronvim/utils/lsp.lua
```diff
@@ -23,17 +23,51 @@ local setup_handlers = user_opts("lsp.setup_handlers", {
...
+      severity_sort = true,
+      float = {
+        focused = false,
+        style = "minimal",
+        border = "rounded",
+        source = "always",
+        header = "",
+        prefix = "",
+      },
+    }),
```
mehalter, line 52: This should just be

```lua
    { 
       virtual_text = false, 
       signs = false, 
       update_in_insert = true, 
       underline = true, 
       severity_sort = true, 
       float = { 
         focused = false, 
         style = "minimal", 
         border = "rounded", 
         source = "always", 
         header = "", 
         prefix = "", 
       }, 
     },
```

lua/astronvim/utils/lsp.lua
```diff
@@ -23,17 +23,51 @@ local setup_handlers = user_opts("lsp.setup_handlers", {
...
```
mehalter, line 68: This Should be

```lua
    { 
       virtual_text = false, 
       signs = { active = signs }, 
       update_in_insert = true, 
       underline = true, 
       severity_sort = true, 
       float = { 
         focused = false, 
         style = "minimal", 
         border = "rounded", 
         source = "always", 
         header = "", 
         prefix = "", 
       }, 
     },
```

lua/astronvim/utils/lsp.lua
```diff
@@ -23,17 +23,51 @@ local setup_handlers = user_opts("lsp.setup_handlers", {
...
+    -- all diagnostics on
+    [3] = astronvim.user_opts("diagnostics", {
```
mehalter, line 70: This line should be

```lua
    astronvim.user_opts("diagnostics", {
```

lua/user_example/init.lua
```diff
@@ -72,7 +71,6 @@ local config = {
 
   -- Diagnostics configuration (for vim.diagnostics.config({...})) when diagnostics are on
   diagnostics = {
-    virtual_text = true,
```
mehalter, line 75: This line shouldn't be deleted

lua/user_example/init.lua
```diff
@@ -72,7 +71,6 @@ local config = {
 
   -- Diagnostics configuration (for vim.diagnostics.config({...})) when diagnostics are on
```
mehalter, line 72: We should probably edit this line to clarify that this is controlling when `vim.g.diagnostics_mode = 3` here

mehalter (changes requested):
Thanks for moving this over! I have made a few code cleanup comments and change suggestions.

mehalter:
@Subjective I went ahead and did this merge to help out a bit because the non-linear git history thing that happened here got super super messy. I made sure to maintain the co-authorship of the commits and also made these small changes as well. Thanks again for the contributions!

mehalter:
Also for reference, here is the new commit: https://github.com/AstroNvim/AstroNvim/commit/1410ee9185a94f96b2aee67e20232b9919e85aab

PR #1592, opened 2023-02-26
feat: allow copying file name and absolute path in neo-tree

Subjective:
This will make it easier to copy file names and directories over to other applications, such as when you want to quickly locate a file in a system file explorer or cd into a directory in the terminal.

### DIFF
--- a/lua/plugins/neo-tree.lua
+++ b/lua/plugins/neo-tree.lua
@@ -83,7 +83,7 @@ return {
           O = "system_open",
           h = "parent_or_close",
           l = "child_or_open",
-          ["Y"] = "copy_file_name",
+          Y = "copy_file_name",
           ["<C-y>"] = "copy_absolute_path",
         },
       },
--- a/lua/plugins/neo-tree.lua
+++ b/lua/plugins/neo-tree.lua
@@ -27,10 +27,7 @@ return {
           state.commands.open(state)
         end
       end,
-      copy_file_name = function(state)
-        local node = state.tree:get_node();
-        vim.fn.setreg('*', node.name, 'c')
-      end,
+      copy_name = function(state) vim.fn.setreg("+", state.tree:get_node().name) end,
       copy_absolute_path = function(state)
         local node = state.tree:get_node();
         local full_path = node.path;
--- a/lua/plugins/neo-tree.lua
+++ b/lua/plugins/neo-tree.lua
@@ -28,11 +28,7 @@ return {
         end
       end,
       copy_name = function(state) vim.fn.setreg("+", state.tree:get_node().name) end,
-      copy_absolute_path = function(state)
-        local node = state.tree:get_node();
-        local full_path = node.path;
-        vim.fn.setreg('*', full_path, 'c')
-      end,
+      copy_path = function(state) vim.fn.setreg("+", state.tree:get_node().path) end,
     }
     local get_icon = require("astronvim.utils").get_icon
     return {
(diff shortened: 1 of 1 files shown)


lua/plugins/neo-tree.lua
```diff
@@ -27,6 +27,15 @@ return {
           state.commands.open(state)
         end
       end,
+      copy_file_name = function(state)
+        local node = state.tree:get_node();
+        vim.fn.setreg('*', node.name, 'c')
+      end,
```
mehalter, line 33: This can be simplified to this, also AstroNvim uses the `+` register by default so that's probably a better default here. Also figured we could simplify the name

```suggestion
      copy_name = function(state) vim.fn.setreg("+", state.tree:get_node().name) end,
```

lua/plugins/neo-tree.lua
```diff
@@ -27,6 +27,15 @@ return {
...
+      copy_absolute_path = function(state)
+        local node = state.tree:get_node();
+        local full_path = node.path;
+        vim.fn.setreg('*', full_path, 'c')
+      end,
```
mehalter, line 38: This can be simplified to this, also AstroNvim uses the `+` register by default so that's probably a better default here. Also figured we could simplify the name

```suggestion
      copy_path = function(state) vim.fn.setreg("+", state.tree:get_node().path) end,
```

lua/plugins/neo-tree.lua
```diff
@@ -74,6 +83,8 @@ return {
           O = "system_open",
           h = "parent_or_close",
           l = "child_or_open",
+          Y = "copy_file_name",
+          ["<C-y>"] = "copy_absolute_path",
```
mehalter, line 87: Update to use the new simplified command names

```suggestion
          Y = "copy_name",
          ["<C-y>"] = "copy_path",
```

mehalter (changes requested):
Thanks for this! I like the idea here, just made a few suggestions for the code. Also for future reference, it's probably a good idea to have `stylua` setup for formatting so that your code can pass the style checking github actions.


Thanks!

Subjective:
Alright, I just accepted the changes.

> Also for future reference, it's probably a good idea to have `stylua` setup for formatting so that your code can pass the style checking github actions.
> 

This may be a bit of a nooby question, but is there any difference between using a dedicated formatter like `stylelua` versus the built-in formatting of `sumneko_lua`?

mehalter (approved):
LGTM

mehalter:
@Subjective `stylua` just had different configuration which we use for this repo. Also the `sumneko_lua` formatting is not super good, `stylua` is just more reliable.

mehalter:
@Subjective I played around with this and did a bit of a rewrite to consolidate this functionality into a single binding for copying various information from a node in neo-tree. Let me know what you think! Commit: https://github.com/AstroNvim/AstroNvim/commit/e283d6749fcd5f2c098751db9e48318205ce3315

Subjective:
@mehalter I think the new changes are really cool! My only qualm is that I typically only use that feature to quickly copy the absolute path, and requiring another input makes doing that ever so slightly slower. `<C-y>` was pretty convenient since I have control binded to my capslock which makes it really easy to hold down ctrl,`<C-h>` into the explorer, select a path to copy, and <C-l> back. I would personally prefer adding that binding back, but I understand if you'd be against doing that out of concern for redundancy.

Also, I'd suggest changing the modifier keys from numbers to letters so you can remain on the home row when copying:
```
p: Path relative to CWD
P: Absolute path
h: Path relative to home
f: Filename
F: Filename w/o extension
e: Extension only
```

mehalter:
Thanks for this feedback @Subjective I think for the base installation having just one binding is good. If you find that you use the absolute path the most you can definitely add a custom binding to your user configuration for `neo-tree`! I just made a commit that moves the modifier keys to use characters for the selection to make this better.

PR #1594, opened 2023-02-27
feat(neotree): add mapping to trash files

Subjective:
This will allow the option of sending files/folders to the system's trash instead of permanently deleting them [(Source)](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/202). However, this only works of the OS supports it — on MacOS, you need to run `brew install trash` to install the CLI command.

Screen Recording 2023-02-26 at 11 49 21 PM

### DIFF
--- a/lua/plugins/neo-tree.lua
+++ b/lua/plugins/neo-tree.lua
@@ -83,7 +83,7 @@ return {
           O = "system_open",
           h = "parent_or_close",
           l = "child_or_open",
-          ["Y"] = "copy_file_name",
+          Y = "copy_file_name",
           ["<C-y>"] = "copy_absolute_path",
         },
       },
--- a/lua/plugins/neo-tree.lua
+++ b/lua/plugins/neo-tree.lua
@@ -27,10 +27,7 @@ return {
           state.commands.open(state)
         end
       end,
-      copy_file_name = function(state)
-        local node = state.tree:get_node();
-        vim.fn.setreg('*', node.name, 'c')
-      end,
+      copy_name = function(state) vim.fn.setreg("+", state.tree:get_node().name) end,
       copy_absolute_path = function(state)
         local node = state.tree:get_node();
         local full_path = node.path;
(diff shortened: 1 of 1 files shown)


mehalter:
This is very system specific so I'm not going to want to merge this in. Feel free to put this in your own user configuration!

Subjective:
> Feel free to put this in your own user configuration!

@mehalter Do you have any suggestions for the optimal way to do this? For now, I just copied my own neo-tree configuration into a separate file in the users directory, but I feel like there must be a more elegant way to do this by just extending the base configuration so that I don't have to manually update that file every time changes are made on the development branch.

mehalter:
@Subjective you could do something like this in your user configuration overriding the plugin definition of `neo-tree` to extend the options:

```lua
    {
      "nvim-neo-tree/neo-tree.nvim",
      opts = function(_, opts)
        -- define new global commands
        local global_commands = {
          trash = function(state)
            local inputs = require "neo-tree.ui.inputs"
            local path = state.tree:get_node().path
            local msg = "Are you sure you want to trash " .. path
            inputs.confirm(msg, function(confirmed)
              if not confirmed then return end

              vim.fn.system { "trash", vim.fn.fnameescape(path) }
              require("neo-tree.sources.manager").refresh(state.name)
            end)
          end,
          trash_visual = function(state, selected_nodes)
            local inputs = require "neo-tree.ui.inputs"

            function GetTableLen(tbl)
              local len = 0
              for _ in pairs(tbl) do
                len = len + 1
              end
              return len
            end

            local count = GetTableLen(selected_nodes)
            local msg = "Are you sure you want to trash " .. count .. " files ?"
            inputs.confirm(msg, function(confirmed)
              if not confirmed then return end
              for _, node in ipairs(selected_nodes) do
                vim.fn.system { "trash", vim.fn.fnameescape(node.path) }
              end
              require("neo-tree.sources.manager").refresh(state.name)
            end)
          end,
        }

        -- apply global commands to each section
        for _, type in ipairs { "filesystem", "buffers", "git_status", "diagnostics" } do
          opts[type].commands = require("astronvim.utils").extend_tbl(opts[type].commands, global_commands)
        end

        -- add new mappings to all windows
        opts.window.mappings.T = "trash"

        -- return new opts
        return opts
      end,
    },
```

PR #1602, opened 2023-03-04
fix(notify): change notification clearing animation to reduce ui clutter

Subjective:
<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/lua/plugins/ui.lua
+++ b/lua/plugins/ui.lua
@@ -47,7 +47,11 @@ return {
   {
     "rcarriga/nvim-notify",
     init = function() require("astronvim.utils").load_plugin_with_func("nvim-notify", vim, "notify") end,
-    opts = { stages = "fade" },
+    opts = {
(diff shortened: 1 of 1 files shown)


luxus:
you can change it for yourself by adding this to your plugins section of your user config.
```lua
  {
    "nvim-notify",
    opts = {
      stages = "slide",
      timeout = 3000,
      max_height = function() return math.floor(vim.o.lines * 0.75) end,
      max_width = function() return math.floor(vim.o.columns * 0.75) end,
    },
  },
```

mehalter:
@Subjective I didn't realize the default animation for `nvim-notify` also moved them up when they went away. I went ahead and removed our conifg and am just using the defaults. I don't want to merge the other stuff that you did here since that should go in your user configuration as @luxus showed here. Thanks for the suggestions though!

PR #1620, opened 2023-03-11
Replace ugly huge chevron arrow in heirline with the small one

stepann0:
Small arrow looks more aesthetic in "heirline.separators"


mehalter:
I think we want to keep the Chevrons for the default setup. Feel free to customize it in your user configuration

PR #1642, opened 2023-03-15
feat(alpha): added startup time to alpha

owittek:
I think adding the startup time to the default configuration might be a nice feature.

The code snippet is based on LazyVim and has been adapted by @luxus & @jay-babu and shared in the [AstroNvim discord](https://discord.com/channels/939594913560031363/1056415608968855572/1078643025472917516).

It looks like this:
image


mehalter (approved):
Thanks! This looks great!

PR #1671, opened 2023-03-18
ci: add spell checker and fix existing typo

hezhizhen:


luxus (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "cSpell.words": ["astronvim"],
+  "Lua.diagnostics.globals": ["vim"]
+}
```
mehalter, line 4: This file should not be committed

typos.toml
```diff
@@ -0,0 +1,6 @@
+# See https://github.com/crate-ci/typos/blob/master/docs/reference.md to configure typos
+[default.extend-words]
+enew = "enew"
+fo = "fo"
```
mehalter, line 4: Could you explain these two word additions?

mehalter (changes requested):

.vscode/settings.json
utkarshgupta137: Awkward... :D

hezhizhen: Oh, sorry...

typos.toml
hezhizhen: As you can see above, they are in the section of `[default.extend-words]`, which means these words are supposed to be correct (they may be not valid English words, but they are valid keywords in a certain programming language, or a commit hash, etc). You can see https://github.com/crate-ci/typos/blob/master/docs/reference.md for more infomation.

mehalter: Yeah, so can you explain why these words shouldn't be considered wrong? Maybe my original question wasn't clear 😅

mehalter: Is this checking the words in the code as well or just the comments?

hezhizhen: Checking all

mehalter: Hm I'll have to think about this and get back to you 🤔 also what about other things we have that aren't English words? Like `ffi`? Why is that not triggering it

hezhizhen: > Yeah, so can you explain why these words shouldn't be considered wrong? Maybe my original question wasn't clear 😅

Oh, I see. `enew` is a vim command and `fo` is a keyboard shortcut, so I think they should be thought as correct.

mehalter: There are a lot more non English words in the code than just those two. I will review this sometime next week and get back to you. Thanks for having this idea :) ideally we would be able to just spell check the comments

hezhizhen: > > Yeah, so can you explain why these words shouldn't be considered wrong? Maybe my original question wasn't clear 😅
> 
> Oh, I see. `enew` is a vim command and `fo` is a keyboard shortcut, so I think they should be thought as correct.

See https://github.com/crate-ci/typos/#false-positives

mehalter: Maintaining these "corrections" will get unruly and unmanageable in the future.

hezhizhen: > Hm I'll have to think about this and get back to you 🤔 also what about other things we have that aren't English words? Like `ffi`? Why is that not triggering it

I haven't read its design docs and I just use it to find out potential misspelling. As far as I see, it's not something like dictionary and list all non-English-valid words.

hezhizhen: > Maintaining these "corrections" will get unruly and unmanageable in the future.

We can skip some certain files in `[files]` section like `mappings.lua` to solve it.

mehalter: There are many many non English words throughout all of the files. I'll get back to you in - a few days.

hezhizhen: > There are many many non English words throughout all of the files. I'll get back to you in - a few days.

And we can see how https://github.com/crate-ci/typos/ works. Maybe it will answer your questions.

mehalter (approved):
I took a look into the architecture of `typos` and it looks good. It does appear that it doesn't actually check against a known dictionary so we shouldn't have to worry about adding every single missing word which is good! Thanks for the recommendation :D

hezhizhen:
@mehalter What have you changed in this PR?

mehalter:
The main thing was just rebasing after other changes got made to `nightly`. I also changed `typos.yml` to `.typos.yml` and removed the explicit definition of the configuration file from the GitHub action workflow since the configuration file can be implicit

hezhizhen:
> The main thing was just rebasing after other changes got made to `nightly`. I also changed `typos.yml` to `.typos.yml` and removed the explicit definition of the configuration file from the GitHub action workflow since the configuration file can be implicit

What's the benefit of renaming `typos.toml` as `.typos.toml`?

mehalter:
> What's the benefit of renaming `typos.toml` as `.typos.toml`?

Clean up the folder a bit by utilizing hidden files. It makes it clearer that dotfiles are there to configure tools that we use in AstroNvim's management and do not actually affect AstroNvim itself. It's more just standard practice.

hezhizhen:
> > What's the benefit of renaming `typos.toml` as `.typos.toml`?
> 
> Clean up the folder a bit by utilizing hidden files. It makes it clearer that dotfiles are there to configure tools that we use in AstroNvim's management and do not actually affect AstroNvim itself. It's more just standard practice.

Thanks for your explanation.

PR #1678, opened 2023-03-21
LSP - Toggle Semantic Token highlights

zbroniszewski:
LSP semantic tokens are not perfect; bugs have been reported. Allow user to toggle semantic token highlights. Also very useful for theme/plugin development.

Neovim 0.9 introduces support for LSP semantic tokens. Users will experience the effect of LSP servers sending incorrect semantic tokens, resulting in incorrect syntax highlighting.

After lengthy discussions with @mehalter, we have seen that it is possible to handle these bugs accordingly, but agreed that we _should not_.
Instead, my suggestion is to provide the ability to disable/toggle semantic tokens as a whole, whether for a buffer, a language server, or globally.
AstroNvim already gives the user the ability to disable semantic tokens globally or per language server via the `lsp.on_attach` user option.
This change addresses toggling semantic token highlighting for a buffer.

Screen Recording 2023-03-20 at 9 26 29 PM

### DIFF
--- a/lua/astronvim/utils/lsp.lua
+++ b/lua/astronvim/utils/lsp.lua
@@ -288,6 +288,22 @@ M.on_attach = function(client, bufnr)
     }
   end
 
+  if capabilities.semanticTokensProvider and vim.lsp["semantic_tokens"] ~= nil then
+    local semantic_highlight_on = true
+    lsp_mappings.n["<leader>uY"] = {
+      function()
+        if semantic_highlight_on then
+          vim.lsp.semantic_tokens.stop(bufnr, client.id)
+          semantic_highlight_on = false
+        else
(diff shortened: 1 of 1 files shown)


mehalter (changes requested):
This technically will have a nasty bug if there are multiple language servers that provide semantic tokens as you will only be able to toggle the most recently attached language server. This should be factored out into a separate function at the root of `utils/lsp.lua` that just takes in a `bufnr`, gets all the clients that have semantic tokens capability enabled (respecting if it has been disabled when they were attached) and then starts/stops all of them based on a vim buffer level variable `vim.b.semantic_highlight_enabled` is similar to what we have elsewhere. Factoring out makes sure we can cache it as well.

PR #1683, opened 2023-03-21
AstroNvim v3.5.0 Nightly

mehalter:


zbroniszewski:
LGTM - I see you changed the meaning of “semantic_tokens_enabled” from “is enabled” to “should be enabled.” Took a second to reason about.

PR #1686, opened 2023-03-22
fix(bootstrap): "improved" config directory detection respects backslash

yehorb:
Looks like the new config directory detection did not take into account that the `vim.fn.stdpath` will return a path with backslashes on Windows. I added the backslash to the pattern.


mehalter (approved):
Ah thanks for catching this! Looks great.

Issue #1710, opened 2023-03-29
Using system_open ("O") in neotree on a directory does not do anything on Windows 10


jkooij:
### Checklist

- [X] Using a stable version of Neovim (i.e. not neovim nightly)
- [X] `:AstroUpdate`
- [X] Restarted AstroNvim

### Operating system/version

Windows 10 Enterprise 64-bit

### Terminal/GUI

nvim-qt

### AstroNvim Health

astronvim: require("astronvim.health").check()
========================================================================
## AstroNvim
  - INFO: AstroNvim Version: v3.5.4
  - INFO: Neovim Version: v0.8.3
  - OK: Using stable Neovim >= 0.8.0
  - OK: `git` is installed: Used for core functionality such as updater and plugin management
  - OK: `explorer` is installed: Used for `gx` mapping for opening files with system opener (Optional)
  - OK: `lazygit` is installed: Used for mappings to pull up git TUI (Optional)
  - OK: `node` is installed: Used for mappings to pull up node REPL (Optional)
  - OK: `gdu` is installed: Used for mappings to pull up disk usage analyzer (Optional)
  - OK: `btm` is installed: Used for mappings to pull up system monitor (Optional)
  - OK: `python` is installed: Used for mappings to pull up python REPL (Optional)

### Describe the bug

In windows, when selecting a directory in Neotree, pressing O for the `system_open` command should open an Explorer window on that directory. Neovim does become defocused, but it appears that explorer does not start, or perhaps it does an immediately closes. In any case, no explorer window is opened.

### Steps to Reproduce

1. Open nvim-qt on windows
2. Change directory to some location with subdirectories
3. Open neotree by pressing `<leader>e`
4. Move cursor on a subdirectory name
5. Press 'O' to call Astronvim's `system_open` command
6. [bug] Nothing happens, no explorer window opens

### Expected behavior

Windows explorer should open after pressing 'O' on a directory in neotree

### Additional Context

~I can provide a pull request, I think.~
I provided pull request #1711 with an easy fix. It should not affect other platforms, but I did not test that.

## Cause
I found the `system_open(path)` function in `lua/astronvim/utils/init.lua`, and with some tests determined that it does correctly detect explorer.exe and uses it as the command for `jobstart`, executing effectively `explorer C:\\Users\\...`. While this command would work in a terminal, it does not seem to work with `jobstart` for some reason.

## Fix
[...]

PR #1711, opened 2023-03-29
fix(utils): system_open in neotree on windows, use "cmd.exe /K explorer" instead of "explorer"

jkooij:
Fix for #1710


mehalter (changes requested):
This is great! Could you make sure to format the code with `stylua` to make the tests pass? I can use squash and merge to fix the conventional commit issue so don't worry about that one.

mehalter (approved):
Thanks!

PR #1730, opened 2023-04-03
refactor: move to using lua_ls for code formatting

mehalter:
Investigating moving from `stylua` for project formatting to just using the Lua Language Server


luxus:
looks like there is a reason why nobody is using it for formating.. so sad :(

PR #1737, opened 2023-04-05
fix(ui): notify when toggling lsp semantic highlights

Subjective:


mehalter (approved):
This is great! Main thing is I do consider this a fix rather than a feature since this should have been there previously! But i can resolve that with squashing. Thanks!

PR #1741, opened 2023-04-05
feat(ui): add <leader>uh binding to toggle foldcolumn

rozhelluk:


lua/astronvim/mappings.lua
```diff
@@ -376,5 +376,6 @@ maps.n["<leader>ut"] = { ui.toggle_tabline, desc = "Toggle tabline" }
 maps.n["<leader>uu"] = { ui.toggle_url_match, desc = "Toggle URL highlight" }
 maps.n["<leader>uw"] = { ui.toggle_wrap, desc = "Toggle wrap" }
 maps.n["<leader>uy"] = { ui.toggle_syntax, desc = "Toggle syntax highlight" }
+if vim.fn.has "nvim-0.9" == 1 then maps.n["<leader>uh"] = { ui.toggle_foldcolumn, desc = "Toggle foldcolumn" } end
```
mehalter, line 379:
```suggestion
maps.n["<leader>uh"] = { ui.toggle_foldcolumn, desc = "Toggle foldcolumn" }
```

This doesn't need to rely on Neovim 0.9. `foldcolumn` is still a feature that could be toggled if the user wants, it just won't be as pretty

lua/astronvim/utils/ui.lua
```diff
@@ -208,4 +208,16 @@ function M.toggle_url_match()
...
 
+--- Toggle foldcolumn=0|1|2
+function M.toggle_foldcolumn()
+  if vim.wo.foldcolumn == "0" then
+    vim.wo.foldcolumn = "1"
+  elseif vim.wo.foldcolumn == "1" then
+    vim.wo.foldcolumn = "2"
+  else
+    vim.wo.foldcolumn = "0"
+  end
```
mehalter, line 219: Should this be between 0/1/2? It probably makes more sense to make the toggle just between 1 and 0. Fold column can be anything between 0 and 9. We can just assume anything other than 1 is user set which is totally fine.

mehalter (changes requested):
This is a great idea! Added a few notes and requests for changes.

mehalter (approved):
This looks great! Thanks!

PR #1768, opened 2023-04-14
fix: fix argument type warning in init.lua when calling utils.notify

tzshen:


mehalter (changes requested):
thanks for catching this! Could you please format with `stylua` to make the tests pass?

mehalter (approved):
Awesome thanks for this!

PR #1780, opened 2023-04-17
3.11.0 Nightly

mehalter:
### TODO

- Wait for Mason.nvim v1.0.0 release to get semantic versioning based plugin updates
- Wait for new Neo-Tree release to update default configuration to be much cleaner (#1781)
- Update `lazy_snapshot`
- Update healthchecks for non-deprecated functions (https://github.com/folke/lazy.nvim/commit/c8cad548950807848de11e3710de2b560758ecb4)


utkarshgupta137:
The treesitter large files check is too conservative. I'm not getting treesitter even for files that have 300 lines (which takes up 10kb). I think it should be of the order of 1MB. I've seen neovim & treesitter handle 10k lines without breaking a sweat.

mehalter:
Ah thanks @utkarshgupta137 ! This should be number of lines and not filesize. I have made a commit to `nightly` to resolve this.

PR #1786, opened 2023-04-18
fix(buffer): fix close without bufdelete

sbeyer:
`<leader>c` with disabled bufdelete plugin leads to error messages:
```
E5108: Error executing lua: .../.config/nvim/lua/astronvim/utils/buffer.lua:72: attempt to concatenate local 'bufnr' (a nil value)                               
stack traceback:
        .../.config/nvim/lua/astronvim/utils/buffer.lua:72: in function 'close'
        .../.config/nvim/lua/astronvim/mappings.lua:49: in function <.../.config/nvim/lua/astronvim/mappings.lua:49>
```

This behavior is fixed by guarding the nil `bufnr` value.


mehalter (approved):
Thanks for catching this!

PR #1803, opened 2023-04-22
fix(neotree): replaces the deprecated tab_labels with sources as recommended by Neotree migrations

akio-hoshino:


mehalter:
This is already resolved on `nightly` which is why this PR cannot be merged. Thanks though!

akio-hoshino:
Oh :)
I should have looked first.
This was my first PR ever to an opensource project.
Thank you for clarifying.

PR #1804, opened 2023-04-23
feat(ui): toggle virtual_lines

zbroniszewski:
Allow user to toggle `virtual_lines` with the same experience as `virtual_text`

For users who do use `virtual_lines`, this feature has been missing from AstroNvim, however it fits nicely into our existing toggles for `virtual_text` with no apparent side-effects.

toggle-virtual-lines

### DIFF
--- a/lua/astronvim/utils/lsp.lua
+++ b/lua/astronvim/utils/lsp.lua
@@ -45,12 +45,12 @@ M.setup_diagnostics = function(signs)
     -- diagnostics off
     [0] = utils.extend_tbl(
       default_diagnostics,
-      { underline = false, virtual_text = false, signs = false, update_in_insert = false }
+      { underline = false, virtual_lines = false, virtual_text = false, signs = false, update_in_insert = false }
     ),
     -- status only
-    utils.extend_tbl(default_diagnostics, { virtual_text = false, signs = false }),
-    -- virtual text off, signs on
-    utils.extend_tbl(default_diagnostics, { virtual_text = false }),
+    utils.extend_tbl(default_diagnostics, { virtual_lines = false, virtual_text = false, signs = false }),
+    -- virtual lines/text off, signs on
+    utils.extend_tbl(default_diagnostics, { virtual_lines = false, virtual_text = false }),
     -- all diagnostics on
     default_diagnostics,
   }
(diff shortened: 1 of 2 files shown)


mehalter (commented):
How does this handle if virtual lines is not an added plugin? Is this added by the plugin LSP Virtual lines plugin?

zbroniszewski:
@mehalter Good point, probably best not to assume here. I actually didn’t see `virtual_lines` in neovim help pages, so I think it’s possible that is somehow added to `vim.diagnostic` by the `lsp_lines` plugin.

I’ll test two things:
- Ensuring `lsp_lines` is disabled/uninstalled first
- Adding a key that definitely does not exist in `vim.diagnostic.config()` and setting it to false to check the behavior, which might help further confirm this is OK for users who do not have `lsp_lines` plugin

Viterkim:
this is a feature from lsp_lines right? I set it up with a custom hotkey, i don't think it's a good idea to add default keybinds for addons which aren't included in astro?

mehalter:
I agree with @Viterkim, I think for AstroNvim's core feature set we shouldn't add things for features that are not provided by the core astronvim plugins

Viterkim:
@zbroniszewski have you setup AstroNvim the idiomatic way? 

https://astronvim.com/Configuration/manage_user_config


https://github.com/AstroNvim/user_example~


It's easier if you have your own changes in the user folder, which gets applied ontop of AstroNvim, that way you don't have your own fork where you have to maintain everything whenever there's an update.

zbroniszewski:
@Viterkim Yes, I've been using this for ~ 2+ years now. My user config files for neovim are here:
https://github.com/zbroniszewski/dotfiles/tree/main/roles/neovim/templates

Sometimes I don't keep track of every single user option and how it works (especially because you have to understand how the underlying plugins use those options), but I've got a good grip.

I might not have explained my last comment well enough. My point was that AstroNvim provides a mapping for diagnostics in general that would be nice to hook into. And _you_ have been forced to create a new mapping to accomplish what _could have_ been hooked into. Further, this should be seen as more broad than diagnostic keymaps in particular. We should be able to hook into as many AstroNvim-provided keymaps as possible for extension as a way of:
- Reducing the number of keymaps a user is forced to create
- Reducing the amount of code a user is required to duplicate/maintain to override an AstroNvim-provided keymap

PR #1816, opened 2023-04-25
chore(gitignore): ignore MacOS System files

vossmalte:
Description:
MacOS System files pollutes `git status` output. Ignoring these files helps (:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ plugin
 ftplugin
 spell
 .luarc.json
+.DS_Store


mehalter:
I personally prefer to keep the `.gitignore` specific to the project itself and Relevant to the tech stack, rather than keeping track of all files that may be present on a person's system. I would recommend setting up a global `.gitignore` on your system that ignores `.DS_Store` on all repos on your machine. It will also save you time needing to add this line to every `.gitignore` you come across.

Thanks for the recommendation but I think in this case I would prefer to keep the `.gitignore` slim and relevant and just filter those files out in PR reviews. Let me know if you disagree!

vossmalte:
Thank you for the comment @mehalter! Personally, I agree with you that setting up a system wide gitignore is better -- on the other hand might be quite the niche trick that most people do not know.

Ergo: 
- keeping files to a minimal scope: definitely yes
- providing project setup files that help most people: yes
-> this PR widens the scope of setup files a little.

I think a larger .gitignore does not hurt the maintainability of a project. It's purpose is to serve the people - not to put a burden to maintain itself.

If you want to keep the repo strictly clean I am happy to close the PR :)

mehalter (approved):
Yeah makes sense to make it easier for people who don't set up their own system files. I will merge this and just go ahead and add a bunch of other system files that we should protect against since I would prefer it be all or nothing. Thanks!

mehalter:
@vossmalte if you are curious, here is the commit I made to continue this: https://github.com/AstroNvim/AstroNvim/commit/6c0223906fea817f51eb9d14ec396b462230e6b4

Let me know if you notice I missed any other system files that might pop up :)

vossmalte:
@mehalter looks good to me! Happy to have contributed 🥳

PR #1859, opened 2023-05-05
feat(dap): add debug Evaluate Expression

chaozwn:
add debug Evaluate Expression for debug.


mehalter (approved):
This looks like a good start! I'll continue modifying this on `nightly`

PR #1868, opened 2023-05-06
fix(plugins): don't load git signs for gitcommit

julianschuler:
Bugs Fixed:
- Fixes `(UNKNOWN PLUGIN): Error executing lua: attempt to call a nil value` error message when writing a git commit message using AstroNvim. This was caused by `gitsigns` being loaded for `gitcommit` files, this PR disables that.

Other:

I don't think there is any use case where `gitsigns` is required for a `gitcommit` file, since the commit message is not in a file tracked by git. Please correct me if I'm wrong.


mehalter (approved):
Thanks! LGTM

PR #1880, opened 2023-05-09
Tabline

lougreenwood:


mehalter:
Also heads up @lougreenwood, I noticed you opened the PR to `main`, I updated this to go to `nightly`

lua/astronvim/utils/status/heirline.lua
mehalter, line 1: The changes to this file should not go into core AstroNvim, this should be put inside of the user configuration using our core API. We are not going to be doing vim-tab based tablines in core AstroNvim but want to enable the user to do it.

We expose the `make_tablist` freely so that you can define it just as you are here in your user configuration.

lua/astronvim/utils/status/provider.lua
mehalter, line 1: The changes in this file are not how this should be tackled. This should be resolved in the component definition to appropriately configure the `bufnr`. It should not be re-evaluated for each individual provider.

lua/plugins/configs/heirline.lua
mehalter, line 1: The changes to this file should be provided in a user configuration and not maintained in core AstroNvim since we will not be providing a vim-tab based tabline out of the box, but simply want to enable the user to do it. Stuff like this and the `maketablist` stuff would be a great addition into the docs and also AstroCommunity.

lua/astronvim/utils/status/heirline.lua
```diff
@@ -47,7 +47,7 @@ M.make_buflist = function(component)
           right = "tabline_bg",
         }
       end,
-      { -- bufferlist
+      {              -- bufferlist
         init = function(self) self.tab_type = M.tab_type(self) end,
```
mehalter, line 51: This is where you should be calculating `self.bufnr` for each tab in the tabline which corresponds to each vim-tab and therefore is where the bufnr should be calculated a single time. It shouldn't be populated as a function through each provider.

lua/astronvim/utils/buffer.lua
mehalter, line 1: This is the one piece in here that looks like it should go into AstroNvim to help enable the user to achieve a vim-tab based tabline. I think this would be a good change to add into AstroCore (here: https://github.com/AstroNvim/astrocore/blob/main/lua/astrocore/buffer.lua) for the upcoming AstroNvim v4 release! We can also just move this single change into v3 to get it in there before the final v3 release and then it will get rebased into v4.

mehalter (changes requested):
I got some time to look into this and this isn't how we are going to want to build this type of thing in AstroNvim. We simply want to make sure that our API can support the user who wants to do this type of thing. There is some code in here that should belong in the user configuration (or AstroCommunity as an entry would be the best)  which will help users add it easily. Stuff like adding highlights for the tabs, defining how to make the tablist, are all things that should be in the user config. Also all of the changes in the providers are unnecessary and should be handled in the component creation.

I do think the `close_tab` function change should actually end up in AstroNvim though!

Could you please revise this PR or make a new one that makes that one change and I can get it released on v3 and/or make a PR to [AstroCore's buffer module](https://github.com/AstroNvim/astrocore/blob/main/lua/astrocore/buffer.lua) and that will make sure it gets into v4

lougreenwood:
Thanks @mehalter - I skimmed through the comments, nothing stood out to me as unreasonable, especially now that `astro-community` is a thing and theoretically this could mostly be bundled up into a recipe in the community.

I'll try to take a look at this in the next week or so and get a PR up.

One other thing, I often notice that my code gets formatted weirdly. AFAIK I only have the `lua` pack installed from the community, I would assume that the code should be auto-formatted correctly if I use this pack?

mehalter:
Yeah it seems that you are formatting things with the lua language server rather than `stylua`. I'm not sure why yours has problems. This should be handled automatically by Neoconf since we have a neoconf config in the root of the project.

PR #1912, opened 2023-05-19
feat: add mapping for lazydocker

MohammadShabaniSBU:


oneslash:
I am not sure if we need this in core? maybe you can add it as a plug-in configuration?

MohammadShabaniSBU:
What do you mean by plug-in configuration?

PR #1922, opened 2023-05-23
PageUp / PageDown support for auto completion

Zeioth:
This is probably behavior the user expects by default.

### DIFF
--- a/lua/plugins/cmp.lua
+++ b/lua/plugins/cmp.lua
@@ -60,6 +60,10 @@ return {
         mapping = {
           ["<Up>"] = cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Select },
           ["<Down>"] = cmp.mapping.select_next_item { behavior = cmp.SelectBehavior.Select },
+          ["<PageUp>"] = cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Select, count=8 },
+          ["<PageDown>"] = cmp.mapping.select_next_item { behavior = cmp.SelectBehavior.Select, count=8 },
+          ["<C-PageUp>"] = cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Select, count=16 },
+          ["<C-PageDown>"] = cmp.mapping.select_next_item { behavior = cmp.SelectBehavior.Select, count=16 },
           ["<C-p>"] = cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Insert },
           ["<C-n>"] = cmp.mapping.select_next_item { behavior = cmp.SelectBehavior.Insert },
           ["<C-k>"] = cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Insert },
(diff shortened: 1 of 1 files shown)


lua/plugins/cmp.lua
```diff
@@ -60,6 +60,12 @@ return {
         mapping = {
           ["<Up>"] = cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Select },
           ["<Down>"] = cmp.mapping.select_next_item { behavior = cmp.SelectBehavior.Select },
+          ["<PageUp>"] = cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Select, count=8 },
+          ["<PageDown>"] = cmp.mapping.select_next_item { behavior = cmp.SelectBehavior.Select, count=8 },
+          ["<C-PageUp>"] = cmp.mapping.select_prev_item { behavior = cmp.SelectBehavior.Select, count=16 },
+          ["<C-PageDown>"] = cmp.mapping.select_next_item { behavior = cmp.SelectBehavior.Select, count=16 },
```
mehalter, line 66:
```suggestion
```

We should remove `<C-PageUp>` and `<C-PageDown>` as this is canonically used for navigating tabs and this is not what people would expect.

mehalter (changes requested):
This is a good idea if people expect it. We should just remove the ctrl bindings since that is not what anyone would expect them to do.

mehalter:
I will actually just go ahead and resolve all of the issues on `nightly` since there are commit message stuff and styling stuff that need to be resolved. Thanks for the suggestions here!

mehalter:
Hm actually @Zeioth I am not sure if these bindings actually do what the user would expect them to. I am going to revert this on `nightly` due to instability. Basically moving up/down by 8 or 16 also wraps around the edge. So if you have a short list or even a long list, going 8/16 down will cause wrapping with whatever is left and makes it really really hard to know what's going on. Let me know what you think of this. But as it is currently this does not seem like a good binding for us to include in core because it can cause some major confusing. If you do want to use this binding you can definitely add it to your user configuration! Sorry for the confusion merging it and then changing my mind. Definitely let me know what you think of the decision.

Zeioth:
Hey! It's totally cool, your decission guys. But it's weird you say instability. What do you mean exactly? Please make sure it is not something else broken on nightly.

[I uploaded a video showing it on my end](https://www.youtube.com/watch?v=LBWjgBG4BM4).

mehalter:
@Zeioth it's not doing a try page down/up operation. If you get to the bottom of the list then it goes back up to the top rather than staying at the bottom which is not expected behavior. Similar with moving up. This is fine for moving up/down one thing but when you do it by `n` spaces you end up cycling by some seemingly non-deterministic amount. If I'm near the bottom of the list and press page down for example. I can end up somewhere between the first 1-8 entries at the top of the list. Which is very very jarring. It would be much better if it hit the bottom of the list and stopped rather than doing this very weird cycling behavior.

Zeioth:
I think the same. Let me check if there is a way to avoid overflowing the list.

Zeioth:
I'm testing it and on my end it doesn't oveflow; When reaching the end of the list, it stops there. Now if you insist a second time, you can oveflow, which is handy because be don't have <HOME> <END> keys in case we want to go back to the top.

Zeioth:
Waiting response from nvim-cmp:
https://github.com/hrsh7th/nvim-cmp/discussions/1587

PR #1935, opened 2023-05-27
chore(mapping): improve mapping descriptions

TheSast:


lua/astronvim/mappings.lua
```diff
@@ -79,23 +79,19 @@ maps.n["<leader>bd"] = {
...
-maps.n["<leader>bsp"] =
-  { function() require("astronvim.utils.buffer").sort "full_path" end, desc = "Sort by full path (buffers)" }
-maps.n["<leader>bsi"] =
-  { function() require("astronvim.utils.buffer").sort "bufnr" end, desc = "Sort by buffer number (buffers)" }
-maps.n["<leader>bsm"] =
-  { function() require("astronvim.utils.buffer").sort "modified" end, desc = "Sort by modification (buffers)" }
+{ function() require("astronvim.utils.buffer").sort "unique_path" end, desc = "By relative path" }
+maps.n["<leader>bsp"] = { function() require("astronvim.utils.buffer").sort "full_path" end, desc = "By full path" }
+maps.n["<leader>bsi"] = { function() require("astronvim.utils.buffer").sort "bufnr" end, desc = "By buffer number" }
+maps.n["<leader>bsm"] = { function() require("astronvim.utils.buffer").sort "modified" end, desc = "By modification" }
```
mehalter, line 94: Should we do these ones? This assumes that the user is only looking at the descriptions in `which-key` in an iterative fashion. The `Sort` is here because if you do `:Telescope keymap` (or whatever it is) you can search through mappings. The description should be understandable by itself and not reliant on the which-key menu title. We should revert this section so that these mappings are still logically searchable.

mehalter (changes requested):
Everything looks great except the changes to the sort descriptions! After those get reverted I can merge this in :) thanks for the contribution!

lua/astronvim/mappings.lua
TheSast: > The description should be understandable by itself and not reliant on the which-key menu title.

I understand the reasoning behind that but not why it should apply to that specific section.

There are other mappings don't make sense outside of context, for example, a good chunk of dap binds: `Step Over (F10)`, `Pause (F6)`, `Start/Continue (F5)`, `Step Into (F11)`, `Restart (C-F5)`, `Terminate Session (S-F5)` (which is also ambiguous to session management), etc.

There are also mappings which lack a proper description, for example, the whichkey ones: operator pending binds, `!`, `"`, `'`, `@`, `>`, `[`, `]`, `v`, `g`, etc. and a bunch of others with an unhelpful `<anonymous>`.

I will still go ahead and revert these ones, but I wanted to ask beforehand: should I partially revert so that they are like the before minus the `(buffers)` part or should I fully revert the commit?

mehalter: I guess that makes sense as well 🤔 maybe we should just leave this as is for now and see if anyone complains. It definitely looks much better.

mehalter (approved):
LGTM

mehalter:
Ah @TheSast can you just make sure to format with `stylua` so that the tests pass?

TheSast:
fixed 👍

PR #1941, opened 2023-05-29
Fix neo-tree empty folder icon with nerdfont v3

Hacksore:
There was a missing key in the config for neo-tree causing empty folders to render the default char shown in the commit below.

https://github.com/nvim-neo-tree/neo-tree.nvim/pull/600/commits/465add19fc88e0257d06da7d5c3ca75c51449f0f#diff-b2cb9452bc9983e124579b119e5d1e01dbe8627f638852ef8ae2ec049c84b920


mehalter:
Thanks!

PR #1942, opened 2023-05-29
 feat(treesitter): Option added → auto_install = true

Zeioth:
This option ensures treesitter always has the necessary language parser when opening a new buffer.

If it is not present, it will install it without asking the user.


mehalter:
This requires the user to have the treesitter cli installed which we do not want to enforce. Without it installed you should expect problems as per the treesitter documentation

PR #1951, opened 2023-06-01
fix(plugins): update nvim-window-picker to v2

mehalter:
Relies on v2 release of [`nvim-window-picker`](https://github.com/s1n7ax/nvim-window-picker)

### DIFF
--- a/lua/lazy_snapshot.lua
+++ b/lua/lazy_snapshot.lua
@@ -43,7 +43,7 @@ return {
   { "rcarriga/nvim-dap-ui", version = "^3", optional = true },
   { "rcarriga/nvim-notify", version = "^3", optional = true },
   { "rebelot/heirline.nvim", commit = "2a151df2dc870e79b138a59ebaaaddf3d1b0d703", optional = true },
-  { "s1n7ax/nvim-window-picker", version = "^1", optional = true },
+  { "s1n7ax/nvim-window-picker", version = "^2", optional = true },
   { "saadparwaiz1/cmp_luasnip", commit = "18095520391186d634a0045dacaa346291096566", optional = true },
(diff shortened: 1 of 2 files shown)


pro465:
hey, so i dont have the time to file a bug report for this, but i can tell you the problem and the fix:
problem: when i click tab for completion, it picks the one under the currently selected item.
fix: `cmp.select_next_item { count = 0 }` [here](https://github.com/AstroNvim/AstroNvim/blob/23bf3b291760686b32d793c0227cee33b85bf873/lua/plugins/cmp.lua#L82)
could you fix it? thanks!

lua/lazy_snapshot.lua
```diff
@@ -41,7 +41,7 @@ return {
   { "rcarriga/nvim-dap-ui", version = "^3", optional = true },
   { "rcarriga/nvim-notify", version = "^3", optional = true },
   { "rebelot/heirline.nvim", commit = "2a151df2dc870e79b138a59ebaaaddf3d1b0d703", optional = true },
-  { "s1n7ax/nvim-window-picker", version = "^1", optional = true },
+  { "s1n7ax/nvim-window-picker", version = "^2", optional = true },
   { "saadparwaiz1/cmp_luasnip", commit = "18095520391186d634a0045dacaa346291096566", optional = true },
   { "stevearc/aerial.nvim", commit = "aff1bb8fecff83d3e3a2d544c4d4e6d65718bd19", optional = true },
   { "stevearc/dressing.nvim", commit = "f19cbd56f7f8cad212c58a7285d09c5d9c273896", optional = true },
```
jay-babu, line 47: line 48, just curious on why pinned at `1.1`?

mehalter: It is pinned to at least v1.1.  so `[1.1,2)` in math notation

PR #1966, opened 2023-06-04
fix: load session with error table index is nil

chaozwn:
when dat.bufnrs[buffname] is nil, cause error table index is nil.


mehalter:
It looks like you don't have the most up to date code. This is fixed in the latest release of AstroNvim

PR #1999, opened 2023-06-15
fix cmp menu width when using Rust

logan-xyz:
fix cmp menu width when using Rust

see : https://github.com/hrsh7th/nvim-cmp/issues/1154


mehalter (approved):
Thanks! I tested this and can't see a difference (not using `rust`) so I will go ahead and merge this. I will make a quick revision on this commit to make the approach a bit more general and improve performance

PR #2000, opened 2023-06-15
Nightly

mehalter:


mehalter:
@logan-xyz Here you can see my revision of your commit based on how we use `lspkind`

PR #2035, opened 2023-06-26
Added explanation for Nerd Fonts to readme

JL102:
The reason for this addition is because it took me a very long time to understand it to begin with.

To new users who never heard of Nerd Fonts, and who may not realize exactly how fonts work on a terminal emulator, they may not understand that you don't actually have to install the Nerd Font on a remote system. It took me a few hours of tinkering to realize this, so the annotation would help future users who end up in my shoes.

The reason for the [1] and [2] is to differentiate between the note related to terminal colors and the note related to nerd fonts. I put the nerd font one above the terminal color one because it appears first in the list.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -67,11 +67,11 @@ AstroNvim is an aesthetic and feature-rich neovim config that is extensible and 
 
 ## ⚡ Requirements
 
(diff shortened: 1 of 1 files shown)


mehalter (approved):
Looks great thanks! I also went ahead and rebased the commits to follow the contribution guidelines so that the tests pass :)

luxus:
why do we recommend iterm2? :(

JL102:
@mehalter Thanks! I didn't notice the contribution guidelines; I'll try and make a habit of checking for contribution guidelines before I submit PRs to new repos.

PR #2036, opened 2023-06-26
fix(lsp) On writing a buffer with no filetype, reload it so LSP kicks in.

Zeioth:
Maybe you can think a more reliable solution but this fixes the issue for now.

## How to test this PR

* Create a new buffer with `:new`
* Write something and save it as `myfile.lua` (or any other filetype)
* Now LSP and heirline LSP indicator will kick in correctly.

### DIFF
--- a/lua/astronvim/autocmds.lua
+++ b/lua/astronvim/autocmds.lua
@@ -59,6 +59,17 @@ autocmd("BufDelete", {
   end,
 })
 
+autocmd("BufWritePre", {
+  desc = "When writing a buffer with no filetype, reload it so lsp kicks in.",
+  group = bufferline_group,
+  callback = function()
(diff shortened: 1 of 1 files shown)


mehalter:
This is not a problem in AstroNvim. Having a buffer with no filetype and writing it to a file does propertly sets the filetype and enables the LSP.

Zeioth:
Just checked on nightly. Doesn't seem to be the case: https://www.youtube.com/watch?v=wQw7StcTb1Y

mehalter:
Ah yeah, looks like there was a difference in my personal configuration that made this work. It shoudl be resolved on nightly now @Zeioth without the need for the new autocmd and also the delayed function of `:edit`. Thanks for pointing out!

Zeioth:
Awesome!

PR #2055, opened 2023-06-28
fix(heirline) VeryLazy instead of BufEnter

Zeioth:
Please feel free to ignore this PR: I've noticed using `BufEnter` instead of `VeryLazy` can cause issues in some border scenarios on component.lsp() where the condition.lsp_attached() report incorrectly. When using a colorcheme other than AstroTheme.

screenshot_2023-06-28_14-10-17_404151422

The image is AstroNvim nightly unchanged, using tokyonight-night.

## How to repro
To reproduce it 100%, open `:term` and run `nvim somefile`. But it can manifest on other scenarios.

My hypothesis is AstroTheme somehow hides stack trace errors from the UI, as it is the only theme I've found where the error doesn't manifest.

### DIFF
--- a/lua/plugins/heirline.lua
+++ b/lua/plugins/heirline.lua
@@ -1,6 +1,6 @@
 return {
   "rebelot/heirline.nvim",
-  event = "BufEnter",
+  event = "VeryLazy",
   opts = function()
     local status = require "astronvim.utils.status"
     return {


mehalter:
Very lazy causes issues. I will investigate different approaches

mehalter:
@Zeioth could you try with the event `UIEnter` and see if you get better results?

mehalter:
@Zeioth I just did some debugging and it looks like this is purely an issue on the current neovim nightly. So this isn't a concern for AstroNvim at the moment. It's a bug upstream

Zeioth:
@mehalter on my end I can replicate on both (arch). I'm outside atm, give me a bit of time to test `UIEnter`

mehalter:
No worries about testing `UIEnter`, it doesn't resolve it. But the error that is coming is coming from core neovim so it's outside of our ability to fix. I'm not concerned about the error because of this. It's coming from calling `get_active_clients`. I am not able to replicate it with stable neovim 0.9.1. I am also only getting this from `tokyonight` and no other colorscheme which also makes me not very concerned about the error. Thanks for reporting though, I will keep it on my radar

mehalter:
I pushed a hacky fix for this that basically delays the activation of the lsp component until our lsp utilities are loaded. Seems to be solid, but very ugly. `VeryLazy` is very much not an option in my mind. It looks soooooooo bad when you open a file directly and causes a ton of lag. If you just do `nvim some_file` you will see the original statusline before the heirline because of the delayed setup

mehalter:
I can confirm that the `VeryLazy` hacky fix does resolve it, but it's just not an option from a usability standpoint

PR #2056, opened 2023-06-28
fix(lsp): remove condition from semantic tokens support check

RayJameson:
Semantic tokens still raise issues with several LS:

https://discord.com/channels/939594913560031363/939857762043695165/1123579379780440094
https://discord.com/channels/939594913560031363/1123569310795825192/1123569310795825192
https://discord.com/channels/939594913560031363/1122910773270818887/1122910773270818887

I am not sure if this will work, because I'm not using any of mentioned LS.


RayJameson:
mehalter said this doesn't work

PR #2067, opened 2023-07-01
feat(neo-tree): add `F` to find in dir neo tree

thuan1412:
Add command `F` to neo-tree in order to `find_files` inside a specific directory. This feature is very helpfull when working in big projects.


mehalter (approved):
Thanks so much! This is a great feature! I will get this merged into nightly and can get it released to stable AstroNvim at the beginning of the week :)

PR #2073, opened 2023-07-03
fix(utils): set which-key title for lsp bindings in visual mode

Subjective:


mehalter:
This isn't necessary, no worries! We don't use `name` key anywhere on AstroNvim, just `desc`

mehalter:
I improved this a bit on nightly. The issue you were fixing will be fixed in AstroNvim v3.30.1

PR #2078, opened 2023-07-04
Only build LuaSnip if NOT on Windows

yehorb:
`make` is usually available on UNIX systems, but the current check sets the `build` property if `win32` IS PRESENT.
This means that the `build` property is present in the case of Windows, and absent in the case of UNIX.
As a Windows user, I get an error every time I update my plugins, and UNIX users don't get the feature they usually can install correctly.
I updated the check to detect if `win32` is NOT `1`, instead of NOT `0`, so that the `build` property is only present on non-Windows systems.

### DIFF
--- a/lua/plugins/cmp.lua
+++ b/lua/plugins/cmp.lua
@@ -1,6 +1,7 @@
 return {
   {
     "L3MON4D3/LuaSnip",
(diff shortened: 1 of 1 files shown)


mehalter (approved):
Thanks for catching this! This is definitely a bug :)

PR #2084, opened 2023-07-05
chore(autocmds): replace `utils.event` with `astroevent` for consistency

Subjective:


mehalter (approved):
Thanks! I went ahead and reworded the commit slightly since this is just a chore and not a fix.

PR #2092, opened 2023-07-05
feat(options): add setting to enable git integration for custom worktrees

Subjective:
Many users (myself included) like tracking their dotfiles in bare git repos which rely on setting a custom worktree to work. Unfortunately, this also means that while editing dotfiles, it's not possible to see git diffs in the gutter or open lazygit in the correct repository as without a `.git` folder it is difficult to know whether a file is being tracked by a bare repo or not.

This PR adds git integration when editing files within custom worktrees, allowing toggleterm and gitsigns gutter highlighting to work as you would expect them to. It also fixes the AstroGitFile event to trigger when a file within a custom worktree is opened, allowing plugins like gitsigns to propely lazy load. 

This functionality is opt-in, so those who don't rely on custom worktrees won't notice anything.

To enable it, set the global option git_worktrees to an array-like table, where each entry represents a worktree with entries `toplevel` and `gitdir`.  This mirrors a similar configuration option in gitsigns (lewis6991/gitsigns.nvim/pull/600). For example, if you have a bare git repo located at ~/.cfg, you can enable git integration for it by setting the following in the AstroNvim global options table:
```lua
git_worktrees = { { toplevel = vim.env.HOME, gitdir = vim.env.HOME .. "/.cfg" } }, -- allow custom git_worktrees to be detected
```

### DIFF
--- a/lua/astronvim/autocmds.lua
+++ b/lua/astronvim/autocmds.lua
@@ -257,7 +257,28 @@ autocmd({ "BufReadPost", "BufNewFile", "BufWritePost" }, {
   callback = function(args)
     if not (vim.fn.expand "%" == "" or vim.api.nvim_get_option_value("buftype", { buf = args.buf }) == "nofile") then
       astroevent "File"
-      if utils.cmd({ "git", "-C", vim.fn.expand "%:p:h", "rev-parse" }, false) then
+      local function in_worktree()
+        for _, worktree in ipairs(vim.g.git_worktrees) do
+          if
+            utils.cmd({
+              "git",
+              "--work-tree",
+              worktree.toplevel,
+              "--git-dir",
+              worktree.gitdir,
+              "ls-files",
+              "--error-unmatch",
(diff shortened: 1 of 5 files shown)


mehalter (approved):
This is awesome! Thanks so much! I'll play around with it and hopefully get a new minor release tagged in the coming days :)

mehalter:
@Subjective I did a slight cleanup on nightly so just let me know if I accidentally broke anything :)

mehalter:
@Subjective do you think you could also make a PR on the documentation to add a new recipe page regarding git working trees? It could give a slight introduction to them and what they are and then describe in more details how to set them up and use them :)

Also for reals, thanks again for this contribution. This is a super great new feature!!!

Subjective:
@mehalter whoops, I didn't see the past few messages for some reason, just opened a [PR](https://github.com/AstroNvim/astronvim.github.io/pull/97) that adds the configuration recipe.

PR #2093, opened 2023-07-05
fix(ui): notifications appearing on top in unintended situations

Subjective:
This fixes an issue where notifications were obscuring UI elements of plugins like [Noice](https://github.com/folke/noice.nvim/blob/main/lua/noice/config/views.lua
).

### DIFF
--- a/lua/plugins/ui.lua
+++ b/lua/plugins/ui.lua
@@ -52,7 +52,7 @@ return {
     init = function() require("astronvim.utils").load_plugin_with_func("nvim-notify", vim, "notify") end,
     opts = {
       on_open = function(win)
-        vim.api.nvim_win_set_config(win, { zindex = 150 })
+        vim.api.nvim_win_set_config(win, { zindex = 175 })
         -- close notification immediately if notifications disabled
         if not vim.g.ui_notifications_enabled then vim.api.nvim_win_close(win, true) end
       end,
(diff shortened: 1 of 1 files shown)


mehalter:
What UI situations would happen in which you would not want the notification to appear on the top? I am not sure I can think of one

mehalter:
A screenshot might be helpful as well showing the unintended behavior

Subjective:
@mehalter The Noice cmdline. From what I can tell, setting the zindex 150 is low enough that it notifications appear above the cmdline while remaining underneath mason and lazy.
Before:
image
After:
image

mehalter:
@Subjective if you set it to `200` will it show the last one on top? If a notification comes up while the command line is open then that would be over, but if you open the command line with a notification open then it would be on top? I'm not sure how ties are resolved. But I think that might be better behavior out of the box if possible

lua/plugins/ui.lua
```diff
@@ -52,7 +52,7 @@ return {
     init = function() require("astronvim.utils").load_plugin_with_func("nvim-notify", vim, "notify") end,
     opts = {
       on_open = function(win)
-        vim.api.nvim_win_set_config(win, { zindex = 1000 })
+        vim.api.nvim_win_set_config(win, { zindex = 150 })
```
mehalter, line 55:
```suggestion
        vim.api.nvim_win_set_config(win, { zindex = 175 })
```

Ok, after some testing and playing around, I think 175 is a good default here. I checked the zindex of our default plugins and this seems to be a good default. If other people want different results then they should override this themselves.

mehalter (changes requested):

mehalter (approved):
Awesome! Thanks for taking the time to open a PR to resolve this :)

PR #2094, opened 2023-07-06
fix(utils): set alpha cursor in a better position

TiagoAraujoDev:
In the `alpha_button` function, its return has an `opts` table that defines the position of the cursor. I changed it to the beginning of the button. Before it was in the middle of the text.

### DIFF
--- a/lua/astronvim/utils/init.lua
+++ b/lua/astronvim/utils/init.lua
@@ -213,7 +213,7 @@ function M.alpha_button(sc, txt)
       position = "center",
       text = txt,
       shortcut = sc,
-      cursor = 5,
+      cursor = -2,
       width = 36,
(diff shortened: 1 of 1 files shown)


mehalter:
Do you have a custom header text? I cannot replicate, when I test on stock AstroNvim on various sized monitors it starts on the top button

mehalter:
ohhhh you mean the beginning of the text I seeee okie!

mehalter (approved):
This is great! Just tested and it looks awesome :)

TiagoAraujoDev:
Thanks, it was my first open source contribution... I'm very happy to be part of AstroNvim!

mehalter:
Glad you are here to contribute 🤗

PR #2104, opened 2023-07-09
fix(utils): allow terminals opened by toggle_term_cmd to dynamically resize

Subjective:
For some reason, terminals opened by [toggleterm.nvim](https://github.com/akinsho/toggleterm.nvim) are not dynamically resized when the terminal window is resized or the font size is changed even though normal floating terminals are. After skimming through the docs, I think may be a bug with toggleterm, but setting `hidden=true` resolves this issue.

As a consequence, this changes the behavior of the `:ToggleTerm` command to toggle the last opened floating terminal, including those opened with a custom command. But honestly, I actually prefer this behavior as I can easily toggle back to my main neovim window from lazygit with `F7` or `<C-'>` without having to constantly quit and reopen it, losing progress. (This is kind of embarassing, but I couldn't figure out a way to switch back as keybindings in lazygit conflict with default neovim bindings). 

This means that to switch back to an existing floating term, you'll need to use a command like `:1ToggleTerm`. But as quitting out of lazygit makes it so that the ToggleTerm command toggles the previous active terminal, I think this is a convenient "side effect".

### DIFF
--- a/lua/astronvim/utils/init.lua
+++ b/lua/astronvim/utils/init.lua
@@ -166,7 +166,7 @@ end
 function M.toggle_term_cmd(opts)
   local terms = astronvim.user_terminals
   -- if a command string is provided, create a basic table for Terminal:new() options
-  if type(opts) == "string" then opts = { cmd = opts, hidden = true } end
+  if type(opts) == "string" then opts = { cmd = opts, hidden = false } end
   local num = vim.v.count > 0 and vim.v.count or 1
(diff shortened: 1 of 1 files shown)


mehalter:
We do not want these bindings to be none hidden terminals. We want to separate the terminals with `<C-'>` and the `<leader>t` mappings to open windows for specific applications. Thanks for the recommendation though! You can set this up in your user configuration :)

PR #2106, opened 2023-07-09
feat: add nvim-dap-virtual-text for nvim dap

chaozwn:
nvim-dap-virtual-text can show variables in line end when you debug.


mehalter:
Thanks for the recommendation. This is a good addition to AstroCommunity. We don't want to add this to core AstroNvim at least for the time being. It can be very disruptive and hard to read while debugging which is why we have left it up to the user to add if they want. The dap ui plugin provides a good overview of this information

PR #2117, opened 2023-07-12
feat(utils): add option to disable telescope integration for lsp keymaps

RayJameson:
I like to use things like `go to reference` with the default window and `bqf` preview, this pr adds handy way to disable telescope integration, which is still enabled by default

image


RayJameson:
wrong branch

PR #2118, opened 2023-07-12
feat(utils): add option to disable telescope integration for lsp keymaps

RayJameson:
I prefer to use things like `go to reference` with the default window and `bqf` preview, this pr adds handy way to disable telescope integration, which is still enabled by default

image

### DIFF
--- a/lua/astronvim/options.lua
+++ b/lua/astronvim/options.lua
@@ -64,6 +64,7 @@ local options = astronvim.user_opts("options", {
     semantic_tokens_enabled = true, -- enable or disable LSP semantic tokens on startup
     ui_notifications_enabled = true, -- disable notifications (TODO: rename to  notifications_enabled in AstroNvim v4)
     git_worktrees = nil, -- enable git integration for detached worktrees (specify a table where each entry is of the form { toplevel = vim.env.HOME, gitdir=vim.env.HOME .. "/.dotfiles" })
+    telescope_lsp_integration_enabled = true, -- enable telescope integration for lsp keymaps
   },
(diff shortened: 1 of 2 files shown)


mehalter:
This is a level of abstraction and misdirection that I would rather not implement when you can already achieve this functionality with the `lsp.mappings` table. Here is an example of a `user/init.lua` file that disables the telescope mappings set there: 

```lua
return {
  lsp = {
    mappings = function(maps)
      if maps.n.gd then maps.n.gd[1] = function() vim.lsp.buf.definition() end end
      if maps.n.gI then maps.n.gI[1] = function() vim.lsp.buf.implementation() end end
      if maps.n.gr then maps.n.gr[1] = function() vim.lsp.buf.references() end end
      if maps.n.gT then maps.n.gT[1] = function() vim.lsp.buf.type_definition() end end
      if maps.n["<leader>lG"] then maps.n["<leader>lG"][1] = function() vim.lsp.buf.workspace_symbol() end end
      if maps.n["<leader>lR"] then maps.n["<leader>lR"][1] = function() vim.lsp.buf.references() end end

      return maps
    end,
  },
}
```

This is a bit more verbose, but is clearer what's going on and doesn't rely on an arbitrary global variable

RayJameson:
I'm gonna be honest, I would probably never would come up with solution you provided, so as regular astronvim user.
Now it makes sense to me, because I didn't consider order in which modules load. I believe, not a lot of users gonna know it either, so I suggested introducing this option

mehalter:
I can add this snippet to the documentation as a recipe :)

RayJameson:
That would be perfect

PR #2119, opened 2023-07-13
feat(mappings): add telescope mapping to fuzzy find in current buffer

Subjective:


lua/astronvim/mappings.lua
```diff
@@ -243,6 +243,8 @@ if is_available "telescope.nvim" then
   maps.n["<leader>gt"] = { function() require("telescope.builtin").git_status() end, desc = "Git status" }
   maps.n["<leader>f<CR>"] = { function() require("telescope.builtin").resume() end, desc = "Resume previous search" }
   maps.n["<leader>f'"] = { function() require("telescope.builtin").marks() end, desc = "Find marks" }
+  maps.n["<leader>f/"] =
+    { function() require("telescope.builtin").current_buffer_fuzzy_find() end, desc = "Find in current buffer" }
```
mehalter, line 247:
```suggestion
    { function() require("telescope.builtin").current_buffer_fuzzy_find() end, desc = "Find words in current buffer" }
```

What do you think of this description? It would align with our other grep like search bindings

mehalter (changes requested):

mehalter (approved):
LGTM!

PR #2121, opened 2023-07-13
v3.33.0 Nightly

mehalter:
For anyone interested in these changes, we are slated to release this on July 17, 2023


lua/astronvim/utils/buffer.lua
```diff
@@ -148,10 +148,12 @@ function M.close(bufnr, force)
         return
       end
     end
-    require("mini.bufremove").delete(bufnr, force)
-  else
-    vim.cmd((force and "bd!" or "confirm bd") .. (bufnr == nil and "" or bufnr))
+    if #vim.fn.tabpagebuflist() > 1 then -- if only one buffer left in tab, run `:bdelete`
```
mehalter, line 151:
```suggestion
    if not M.is_valid(bufnr) or #vim.t.bufs > 1 then
```

This seems like more the intended behavior we would want in this case

PR #2124, opened 2023-07-14
fix(mappings): use git root of current file for git telescope mappings

Subjective:
This makes it so that the git root used for telescope commands is determined based on the file you are in when calling it. This prevents instances where the wrong git commits/branches are brought up. For example, if you have a project open, but you decide to open up a single file from another git repo (say your `.zshrc`), `require("telescope.builtin").git_commits()` will bring up the commit history for the original project, while `require("telescope.builtin").git_bcommits()` won't even work.


mehalter (approved):
LGTM!

PR #2125, opened 2023-07-15
chore(alpha): remove unecessary alpha startup autocmd

Subjective:
This fixes an issue where the FileType autocmd is not emitted by Alpha, resulting in plugins like `mini.indentscope` being enabled on the alpha.

### DIFF
--- a/lua/astronvim/autocmds.lua
+++ b/lua/astronvim/autocmds.lua
@@ -183,24 +183,6 @@ if is_available "alpha-nvim" then
       end
     end,
   })
-  autocmd("VimEnter", {
-    desc = "Start Alpha when vim is opened with no arguments",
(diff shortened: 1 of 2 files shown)


mehalter:
This is not something we want to do. Thanks for the recommendation though!

mehalter:
Feel free to implement this in your own user configuration though if you want this behavior

Subjective:
Just curious, what's the rationale behind having a custom autocmd to load alpha as opposed to doing it via lazy? It seems like alpha will autostart on its own in the appropriate situations (e.g. when no arguments are called).

mehalter:
If the user opens a file directly then alpha will not load on startup

Subjective:
I'm honestly amazed by the amount of foresight to shave off that 1ms startup time 😭

PR #2126, opened 2023-07-15
fix: closing last buffer in a tabpage closes the tab

teddywaweru:
Previously, closing a the last buffer resulted in the next loaded buffer in the jumplist being displayed in the current tabpage window. The fix checks if there are less than 2 buffers in the tab. If yes, save the current buffer_number, close the tab, and proceed to close the buffer(which remains hidden at this point). The next tab's window is selected, depending on the jumplist.

### DIFF
--- a/lua/astronvim/utils/buffer.lua
+++ b/lua/astronvim/utils/buffer.lua
@@ -148,6 +148,10 @@ function M.close(bufnr, force)
         return
       end
     end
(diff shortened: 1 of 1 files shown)


mehalter:
This is actually not the appropriate fix here because this doesn't include all buffers. This doesn't account for unlisted buffers and this will cause issues. The correct fox will be reporting this bug upstream to be resolved

PR #2127, opened 2023-07-16
feat: remove the ctrl+hjkl mappings in lazygit to prevent conflicts.

chaozwn:
we can use `q` esc lazygit, and don't need `ctrl+hjkl` in lazygit. and when we in lazygit commits module. `ctrl+hjkl` is conflict with our keymapping.

### DIFF
--- a/lua/astronvim/mappings.lua
+++ b/lua/astronvim/mappings.lua
@@ -316,7 +316,22 @@ if is_available "toggleterm.nvim" then
       function()
         local worktree = require("astronvim.utils.git").file_worktree()
         local flags = worktree and (" --work-tree=%s --git-dir=%s"):format(worktree.toplevel, worktree.gitdir) or ""
-        utils.toggle_term_cmd("lazygit " .. flags)
+        utils.toggle_term_cmd {
+          cmd = "lazygit " .. flags,
+          hidden = true,
+          on_open = function()
+            vim.api.nvim_del_keymap("t", "<C-h>")
+            vim.api.nvim_del_keymap("t", "<C-j>")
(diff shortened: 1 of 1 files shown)


mehalter:
You can add this to your user configuration. This change in behavior is a bit jarring and not intuitive and will take away from the user experience especially in cases where people use it to switch windows for their multiplexer. Thanks for the recommendation though!

mehalter:
As another suggestion, you can also modify your lazygit configuration to change the bindings in lazygit as well to avoid the conflicts

chaozwn:
> As another suggestion, you can also modify your lazygit configuration to change the bindings in lazygit as well to avoid the conflicts

I don't quite agree with that. Modifying the config of lazygit does have a certain learning curve for users. Using Ctrl+hjkl to switch between float windows is actually an incorrect approach. For example, I'm used to using Ctrl+h to delete in insert mode, which would force me to use the backspace key in this case.

mehalter:
you don't have to open lazy_git in a floating window. If the user changes their default position to vertical/horizontal split (which many users do) then those bindings actually won't make floating terminals. They will make splits. In which case those bindings are completely necessary. Feel free to disagree with these opinions set by AstroNvim and change them in your user configuration :)

chaozwn:
> you don't have to open lazygit in a floating window. If the user changes their default position to vertical/horizontal split (which many users do) then those bindings actually won't make floating terminals. They will make splits. In which case those bindings are completely necessary. Feel free to disagree with these opinions set by AstroNvim and change them in your user configuration :)

oh, i got it.

PR #2155, opened 2023-07-21
fix(ui): notify when toggling url highlighting

Subjective:
This PR adds a notification when toggling url highlighting to be consistent with the other url toggles.


Subjective:
On a related note, might there be any way to improve the performance of url highlighting? Having it enabled makes highlighting quite a bit slower, which is especially apparent using leap with macros:

Screen Recording 2023-07-21 at 2 47 10 PM

Subjective:
Though I suppose a better solution might just be to not use highlighting when executing macros with leap, maybe this can be done with an autocmd

mehalter (approved):
Thanks!

mehalter:
This actually was merged to `nightly`. I resolved the incorrect merge base with rebasing because I caught it late. Sorry!

PR #2156, opened 2023-07-22
chore(status): remove unecessary callback for `macro_recording`

Subjective:
See https://github.com/rebelot/heirline.nvim/pull/133

### DIFF
--- a/lua/astronvim/utils/status/component.lua
+++ b/lua/astronvim/utils/status/component.lua
@@ -110,7 +110,12 @@ function M.cmd_info(opts)
     macro_recording = {
       icon = { kind = "MacroRecording", padding = { right = 1 } },
       condition = condition.is_macro_recording,
-      update = { "RecordingEnter", "RecordingLeave" },
+      update = {
+        "RecordingEnter",
+        "RecordingLeave",
+        -- TODO: remove when dropping support for Neovim v0.8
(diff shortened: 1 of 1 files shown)


mehalter (changes requested):
We should revert this back to what it is and just put a TODO comment to remove it when dropping support for Neovim 0.8

lua/astronvim/utils/status/component.lua
```diff
@@ -110,11 +110,7 @@ function M.cmd_info(opts)
     macro_recording = {
       icon = { kind = "MacroRecording", padding = { right = 1 } },
       condition = condition.is_macro_recording,
-      update = {
-        "RecordingEnter",
-        "RecordingLeave",
-        callback = vim.schedule_wrap(function() vim.cmd.redrawstatus() end),
-      },
+      update = { "RecordingEnter", "RecordingLeave" },
```
mehalter, line 113:
```suggestion
      update = {
        "RecordingEnter",
        "RecordingLeave",
        -- TODO: remove when dropping support for Neovim v0.8
        callback = vim.fn.has "nvim-0.9" == 0 and vim.schedule_wrap(function() vim.cmd.redrawstatus() end) or nil,
      },
```

mehalter (changes requested):
Made suggestion to not remove the callback but make it conditional and also add todo to remove it later

mehalter (approved):
LGTM!

PR #2159, opened 2023-07-22
feat(mappings): add visual selection git commits telescope mapping

Subjective:
It seems like opening the picker doesn't exit visual mode automatically, so this hack is needed. Also, should the other selection mode (`v`) mappings be changed to visual mode (`x`) mappings?

Edit: the hack is no longer needed as of https://github.com/nvim-telescope/telescope.nvim/commit/1228f3b15ca3d9b95dcb92efda6a3448871030bd

### DIFF
--- a/lua/astronvim/mappings.lua
+++ b/lua/astronvim/mappings.lua
@@ -246,7 +246,7 @@ if is_available "telescope.nvim" then
     desc = "Git commits (current file)",
   }
   maps.x["<leader>gc"] = {
-    '<cmd>lua require("telescope.builtin").git_bcommits_range { use_file_path = true }<cr><esc>',
+    function() require("telescope.builtin").git_bcommits_range { use_file_path = true } end,
     desc = "Git commits (selected range)",
   }
   maps.n["<leader>gt"] =
--- a/lua/astronvim/mappings.lua
+++ b/lua/astronvim/mappings.lua
@@ -245,6 +245,10 @@ if is_available "telescope.nvim" then
     function() require("telescope.builtin").git_bcommits { use_file_path = true } end,
     desc = "Git commits (current file)",
   }
+  maps.x["<leader>gc"] = {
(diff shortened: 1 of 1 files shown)


lua/astronvim/mappings.lua
```diff
@@ -245,6 +245,10 @@ if is_available "telescope.nvim" then
     function() require("telescope.builtin").git_bcommits { use_file_path = true } end,
     desc = "Git commits (current file)",
   }
+  maps.x["<leader>gc"] = {
```
mehalter, line 248: Mode `x` includes normal mode. So we should remove the `n` mode and make this just a single mapping. We also need to fix the messages to the the same

mehalter (changes requested):

mehalter:
Oh I just saw your comment on the PR. `x` mode and `v` mode are completely different. As in `x` includes normal mode. If there is a bug in telescope, that's unsupported from our code and not something I'm worried about. We don't need to change any of our `v` mode mappings to `x`

lua/astronvim/mappings.lua
Subjective: Shouldn't `x` be only visual mode while `v` includes visual and selection mode? I'm not sure if the other mappings should apply to selection mode as well.

Subjective:
> Oh I just saw your comment on the PR. `x` mode and `v` mode are completely different. As in `x` includes normal mode. If there is a bug in telescope, that's unsupported from our code and not something I'm worried about. We don't need to change any of our `v` mode mappings to `x`

Shouldn't x be only visual mode while v includes visual and selection mode? I'm not sure if the other mappings should apply to selection mode as well.

mehalter:
Ah you are correct. Also I've thought about this and I don't think we need this to be a default mappings. We don't necessarily want to map every default telescope picker in core AstroNvim. Feel free to add more mappings to your user configuration

PR #2161, opened 2023-07-23
fix(status): always show correct filename for files without a name

Subjective:
This PR does two things: 

1. It changes the default fallback to display noname buffers in heirline as "[No Name]"  instead of "Empty" (which imo doesn't really make sense once the buffer has been modified) to align with default vim naming.
2. It also fixes an issue where when a modifier (other than ":t") is applied to the `file_info` component, the fallback is not always shown, for example with the following custom component:
```lua
status.component.file_info {
  filename = { modify = ":~:." },
},
```

The issue is that applying custom modifiers to `filename` for a `[No Name]` file sets it from an empty string to a nonempty string. This PR resolves this by performing the fallback check on the original path before it was modified.

Here's a screenshot showing the changes:

image

### DIFF
--- a/lua/astronvim/utils/status/provider.lua
+++ b/lua/astronvim/utils/status/provider.lua
@@ -293,8 +293,9 @@ function M.filename(opts)
     modify = ":t",
   }, opts)
   return function(self)
-    local filename = vim.fn.fnamemodify(opts.fname(self and self.bufnr or 0), opts.modify)
-    return status_utils.stylize((filename == "" and opts.fallback or filename), opts)
+    local path = opts.fname(self and self.bufnr or 0)
+    local filename = vim.fn.fnamemodify(path, opts.modify)
+    return status_utils.stylize((path == "" and opts.fallback or filename), opts)
   end
 end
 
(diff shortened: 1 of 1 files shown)


lua/astronvim/utils/status/provider.lua
```diff
@@ -288,13 +288,14 @@ end
 -- @see astronvim.utils.status.utils.stylize
 function M.filename(opts)
   opts = extend_tbl({
-    fallback = "Empty",
+    fallback = "[No Name]",
```
mehalter, line 291:
```suggestion
    fallback = "Empty",
```

Please revert this part as I would like to leave this as "Empty". It's not saying the buffer is empty, but the name is empty. Also it saves room. You can override this in your user configuration if you prefer "[No Name]"

mehalter (changes requested):
Thanks! Just small part to revert

mehalter (approved):
Went ahead and applied the changes. Thanks!

Subjective:
@mehalter Do you think it would make sense to rename "Empty" to "Untitled" since that's what Vim calls the noname buffer when closing in the confirm prompt?

image

mehalter:
Sure, I pushed that. Also you can definitely change this in your user configuration if you have differnt strings you want this to be in the future :)

PR #2162, opened 2023-07-23
fix(ui): improve toggle syntax highlighting

Subjective:
Currently, `toggle_syntax_highlighting` is rather janky. It completely breaks when `vim.g.ui_notifications_enabled=true` (prolly something to do with syntax highlighting within the notify buffer) and doesn't always work as expected between multiple buffers.

This PR does 3 things:
1. It fixes syntax highlighting when ui notifications are enabled
2. It fixes the inconsistent toggling behavior between buffers by enabling and disabling syntax locally instead of globally (TS highlighting was already buffer-local before, this just makes native syntax highlighting buffer-local as well)
3. It also disables/enables semantic token highlighting when appropriate

### DIFF
--- a/lua/astronvim/utils/ui.lua
+++ b/lua/astronvim/utils/ui.lua
@@ -81,12 +81,15 @@ end
 
 --- Toggle buffer semantic token highlighting for all language servers that support it
 ---@param bufnr? number the buffer to toggle the clients on
-function M.toggle_buffer_semantic_tokens(bufnr)
+--- @param silent? boolean if true then don't sent a notification
+function M.toggle_buffer_semantic_tokens(bufnr, silent)
   vim.b.semantic_tokens_enabled = not vim.b.semantic_tokens_enabled
   for _, client in ipairs(vim.lsp.get_active_clients()) do
     if client.server_capabilities.semanticTokensProvider then
       vim.lsp.semantic_tokens[vim.b.semantic_tokens_enabled and "start" or "stop"](bufnr or 0, client.id)
-      notify(string.format("Buffer lsp semantic highlighting %s", bool2str(vim.b.semantic_tokens_enabled)))
+      if not silent then
+        notify(string.format("Buffer lsp semantic highlighting %s", bool2str(vim.b.semantic_tokens_enabled)))
+      end
     end
   end
 end
@@ -206,9 +209,11 @@ function M.toggle_syntax()
   if vim.bo[buf].syntax == "off" then
     if ts_avail and parsers.has_parser() then vim.cmd.TSBufEnable "highlight" end
     vim.cmd.setlocal "syntax=on"
+    if not vim.b.semantic_tokens_enabled then M.toggle_buffer_semantic_tokens(buf, true) end
   else
     if ts_avail and parsers.has_parser() then vim.cmd.TSBufDisable "highlight" end
     vim.cmd.setlocal "syntax=off"
+    if vim.b.semantic_tokens_enabled then M.toggle_buffer_semantic_tokens(buf, true) end
   end
   notify(string.format("syntax %s", bool2str(vim.bo[buf].syntax ~= "off")))
 end
--- a/lua/astronvim/utils/ui.lua
+++ b/lua/astronvim/utils/ui.lua
@@ -203,9 +203,10 @@ function M.toggle_wrap()
 end
 
 --- Toggle syntax highlighting and treesitter
-function M.toggle_syntax()
+---@param bufnr? number the buffer to toggle syntax on
+function M.toggle_buffer_syntax(bufnr)
   local ts_avail, parsers = pcall(require, "nvim-treesitter.parsers")
-  local buf = vim.api.nvim_win_get_buf(0)
+  local buf = bufnr or vim.api.nvim_win_get_buf(0)
   if vim.bo[buf].syntax == "off" then
     if ts_avail and parsers.has_parser() then vim.cmd.TSBufEnable "highlight" end
     vim.cmd.setlocal "syntax=on"
--- a/lua/astronvim/utils/ui.lua
+++ b/lua/astronvim/utils/ui.lua
@@ -204,7 +204,7 @@ end
 
 --- Toggle syntax highlighting and treesitter
 ---@param bufnr? number the buffer to toggle syntax on
-function M.toggle_buffer_syntax(bufnr)
+function M.toggle_syntax(bufnr)
   local ts_avail, parsers = pcall(require, "nvim-treesitter.parsers")
   local buf = bufnr or vim.api.nvim_win_get_buf(0)
   if vim.bo[buf].syntax == "off" then
--- a/lua/astronvim/utils/ui.lua
+++ b/lua/astronvim/utils/ui.lua
@@ -202,14 +202,15 @@ end
 --- Toggle syntax highlighting and treesitter
 function M.toggle_syntax()
   local ts_avail, parsers = pcall(require, "nvim-treesitter.parsers")
-  if vim.g.syntax_on then -- global var for on//off
-    if ts_avail and parsers.has_parser() then vim.cmd.TSBufDisable "highlight" end
-    vim.cmd.syntax "off" -- set vim.g.syntax_on = false
-  else
+  local buf = vim.api.nvim_win_get_buf(0)
+  if vim.bo[buf].syntax == "off" then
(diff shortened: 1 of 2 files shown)


lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
 --- Toggle syntax highlighting and treesitter
 function M.toggle_syntax()
```
mehalter, line 206: This Shou take `bufnr` as a parameter just like the other buffer local toggles. This should default to current buffer if none is provided

lua/astronvim/utils/ui.lua
```diff
@@ -81,12 +81,15 @@ end
 
 --- Toggle buffer semantic token highlighting for all language servers that support it
 ---@param bufnr? number the buffer to toggle the clients on
-function M.toggle_buffer_semantic_tokens(bufnr)
+--- @param silent? boolean if true then don't sent a notification
+function M.toggle_buffer_semantic_tokens(bufnr, silent)
```
mehalter, line 85: I like the idea of adding `silent`, but we should do this for all of the functions at once and not start misaligning these function definitions

mehalter (changes requested):
Just a couple small change requests

lua/astronvim/utils/ui.lua
mehalter, line 206:
```suggestion
function M.toggle_syntax(bufnr)
bufnr = bufnr or 0
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
 --- Toggle syntax highlighting and treesitter
```
mehalter, line 205:
```suggestion
--- Toggle syntax highlighting and treesitter
---@param bufnr? number the buffer to syntax highlighting on
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
 function M.toggle_syntax()
   local ts_avail, parsers = pcall(require, "nvim-treesitter.parsers")
-  if vim.g.syntax_on then -- global var for on//off
-    if ts_avail and parsers.has_parser() then vim.cmd.TSBufDisable "highlight" end
-    vim.cmd.syntax "off" -- set vim.g.syntax_on = false
-  else
+  local buf = vim.api.nvim_win_get_buf(0)
```
mehalter, line 208:
```suggestion
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
+  if vim.bo[buf].syntax == "off" then
```
mehalter, line 209:
```suggestion
  if vim.bo[bufnr].syntax == "off" then
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
     if ts_avail and parsers.has_parser() then vim.cmd.TSBufEnable "highlight" end
```
mehalter, line 210:
```suggestion
    if ts_avail and parsers.has_parser() then vim.treesitter.start(bufnr) end
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
-    vim.cmd.syntax "on" -- set vim.g.syntax_on = true
+    vim.cmd.setlocal "syntax=on"
```
mehalter, line 211:
```suggestion
    vim.bo[bufnr].syntax = "on"
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
+    if not vim.b.semantic_tokens_enabled then M.toggle_buffer_semantic_tokens(buf, true) end
+  else
+    if ts_avail and parsers.has_parser() then vim.cmd.TSBufDisable "highlight" end
```
mehalter, line 214:
```suggestion
    if ts_avail and parsers.has_parser() then vim.treesitter.stop(bufnr) end
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
+    vim.cmd.setlocal "syntax=off"
```
mehalter, line 215:
```suggestion
    vim.bo[bufnr].syntax = "off"
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
-  if vim.g.syntax_on then -- global var for on//off
-    if ts_avail and parsers.has_parser() then vim.cmd.TSBufDisable "highlight" end
-    vim.cmd.syntax "off" -- set vim.g.syntax_on = false
-  else
+  local buf = vim.api.nvim_win_get_buf(0)
+  if vim.bo[buf].syntax == "off" then
     if ts_avail and parsers.has_parser() then vim.cmd.TSBufEnable "highlight" end
-    vim.cmd.syntax "on" -- set vim.g.syntax_on = true
+    vim.cmd.setlocal "syntax=on"
+    if not vim.b.semantic_tokens_enabled then M.toggle_buffer_semantic_tokens(buf, true) end
```
mehalter, line 212:
```suggestion
    if not vim.b[bufnr].semantic_tokens_enabled then M.toggle_buffer_semantic_tokens(bufnr, true) end
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
+  else
+    if ts_avail and parsers.has_parser() then vim.cmd.TSBufDisable "highlight" end
+    vim.cmd.setlocal "syntax=off"
+    if vim.b.semantic_tokens_enabled then M.toggle_buffer_semantic_tokens(buf, true) end
```
mehalter, line 216:
```suggestion
    if vim.b[bufnr].semantic_tokens_enabled then M.toggle_buffer_semantic_tokens(bufnr, true) end
```

lua/astronvim/utils/ui.lua
```diff
@@ -202,14 +205,17 @@ end
...
   end
-  notify(string.format("syntax %s", bool2str(vim.g.syntax_on)))
+  notify(string.format("syntax %s", bool2str(vim.bo[buf].syntax ~= "off")))
```
mehalter, line 218:
```suggestion
  notify(string.format("syntax %s", vim.bo[buf].syntax))
```

This is actually already going to be `on` or `off`

mehalter (commented):
Some more specific recommendations

mehalter:
@Subjective we shouldn't rename the function. Or at least copy it to the old name too just because this is considered breaking

lua/astronvim/utils/ui.lua
```diff
@@ -200,16 +203,20 @@ function M.toggle_wrap()
...
 
 --- Toggle syntax highlighting and treesitter
-function M.toggle_syntax()
+---@param bufnr? number the buffer to toggle syntax on
+function M.toggle_syntax(bufnr)
   local ts_avail, parsers = pcall(require, "nvim-treesitter.parsers")
-  if vim.g.syntax_on then -- global var for on//off
-    if ts_avail and parsers.has_parser() then vim.cmd.TSBufDisable "highlight" end
-    vim.cmd.syntax "off" -- set vim.g.syntax_on = false
+  bufnr = bufnr or vim.api.nvim_win_get_buf(0)
```
mehalter, line 209:
```suggestion
  bufnr = bufnr or 0
```

All of these functions/options handle 0 already. No need for us to expand it

lua/astronvim/utils/ui.lua
Subjective: Is that really necessary given that there's already a global check for `ui_notifications_enabled`? I only added it for `toggle_buffer_semantic_tokens` since it made sense for that specific situation.

mehalter: It's good for these functions to have silent options if people want to script with this API for whatever reason. Seems to be a good option in this API

mehalter (changes requested):

mehalter (approved):
Looks good, I'll finish this up on `nightly`

lua/astronvim/utils/ui.lua
mehalter: also @Subjective I made a PR to neovim core to fix this. We do have to expand because of a bug in `vim.treesitter.stop` not aligning APIs: https://github.com/neovim/neovim/pull/24450

PR #2163, opened 2023-07-23
feat(ui): add optional silent argument to ui toggle functions

Subjective:


mehalter:
Ah I already added this actually on `nightly` a few minutes ago! Sorry!

PR #2164, opened 2023-07-23
fix(ui): duplicate lsp semantic token toggle notification

Subjective:


mehalter:
Hm, this isn't the correct approach. This actually needs to be fixed in a bigger way. I will updates to this. Also I'm not sure if I understand the PR title since the changes made here don't actually do what the PR title says.

Subjective:
> Also I'm not sure if I understand the PR title since the changes made here don't actually do what the PR title says.

Ah, you're right, whoops

PR #2165, opened 2023-07-23
fix(ui): don't check every lsp client when toggling semantic tokens

Subjective:
This *actually* fixes the duplicate messages when toggling semantic token highlighting.

Screen Recording 2023-07-23 at 8 32 38 AM


mehalter (approved):
LGTM! This doesn't fully fix this, but I can resolve append this

PR #2167, opened 2023-07-24
fix(autocmds): improve `AstroColorScheme` autocmds

Subjective:
Summary: 

1. Don't emit AstroColorScheme twice on startup
2. Fixes a weird issue where heirline hl groups are incorectly set (it happens each time lazy auto installs a plugin on startup and other times rather inconsistently). I honestly thought 1) would do the trick, but apparently not 🤷, lmk if there's a better way of going about things

image


mehalter:
This actually just completely breaks our highlight table 😂 we shut discuss probably on discord or something what the actual bug is and how to proceed rather than merging this which basically just disables our highlight modifications 😂

PR #2176, opened 2023-07-26
feat(options): add option to automatically remove trailing whitespace

Subjective:
Feel free to reject if this is out of scope


Subjective:
On second thought, maybe we shouldn't notify if the formatter can already remove whitespace. Not really sure how to check that though.

mehalter:
This should be handled by the formatter or the user can add this to their user configuration

PR #2188, opened 2023-07-31
feat(neo-tree): add mappings to navigate neo-tree filter popup

Subjective:
Sets `<c-j>`/`<c-/k>` to navigate up/down neo-tree filter results to align with telescope and cmp bindings

### DIFF
--- a/lua/plugins/neo-tree.lua
+++ b/lua/plugins/neo-tree.lua
@@ -126,6 +126,10 @@ return {
           l = "child_or_open",
           o = "open",
         },
+        fuzzy_finder_mappings = { -- define keymaps for filter popup window in fuzzy_finder_mode
+          ["<C-j>"] = "move_cursor_down",
+          ["<C-k>"] = "move_cursor_up",
+        },
       },
       filesystem = {
         follow_current_file = { enabled = true },


mehalter:
We don't want to use this because this conflicts with window navigation which there can be a window below neo-tree

Subjective:
@mehalter this is an insert mode mapping, so I don't think it should affect window navigation

mehalter:
Ah I see I see, I'll take a look tomorrow!

Subjective:
In case you don't have access to your setup, this is what it looks like: Screen Recording 2023-07-30 at 5 30 58 PM

mehalter:
Okie! I'll have access to a computer tomorrow and I'll get this merged in and stuff

mehalter (dismissed):
LGTM

mehalter:
Oh actually after investigating @Subjective there are plenty of default mappings. I think it's a good idea for us to just rely on those maybe and leave it up to the user to extend these? What do you think?

Subjective:
Yeah, I noticed that too. I agree that it certainly feels redundant to add duplicate bindings. But at the same time, `<c-n>`/`<c-p>`  are not the most obvious bindings given AstroNvim's defaults for `telescope` and `nvim-cmp`, so a new user might be confused about how to navigate the filtered results, at least not without referencing the docs. `nvim-cmp` has the same duplicate bindings, so I personally don't think it's all that inconsistent to add these, especially since they can be overridden anyways.

Of course, that's just my take on this, as always the final decision up to you :)

mehalter (approved):
LGTM!

mehalter:
After thinking about this, I think having all 3 sets of movement commands to align with `cmp` is a good default!

mehalter:
I also reworded the commit because I think I'm going to consider this a fix because we should have had that aligned before :)

PR #2230, opened 2023-08-20
chore(options): remove "How-to disable mouse" menu item

Subjective:
## 📑 Description

- Removes "How-to disable mouse" from right click menu

## ℹ Additional Information

**Before**:
image

**After**:
image


mehalter:
We don't do anything custom at the moment for the right click menu so I think we actually want to keep this default. Feel free to add to your own configuration!

PR #2233, opened 2023-08-23
fix(neo-tree)!: conflict with new "Order by" keymaps

Subjective:
## 📑 Description

Removes the "o" keymap as there is a conflict with Neotree's new default ["Order by" keymaps]
(https://github.com/nvim-neo-tree/neo-tree.nvim/commit/70ab62c8cc5d039766c039597c7c33a76d74fd11). Hopefully shouldn't be too much of a breaking change since there are several other existing keymaps that do the same thing.

## ℹ Additional Information
image

### DIFF
--- a/lua/astronvim/plugins/neo-tree.lua
+++ b/lua/astronvim/plugins/neo-tree.lua
@@ -171,7 +171,6 @@ return {
           Y = "copy_selector",
           h = "parent_or_close",
           l = "child_or_open",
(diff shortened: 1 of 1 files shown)


mehalter:
There is no such thing as "too much" of a breaking change. This should just go on v4 release since it is categorically not allowed in v3. Feel free to do this in your own configuration in the meantime

PR #2236, opened 2023-08-24
chore(heirline): add tabline padding to undotree

Subjective:
## 📑 Description

Adds tabline padding for undotree.

At first I thought it would be best to leave this in a personal config since undo tree is not included in AstroNvim, but unfortunately doing so is not that simple and results in the mental overhead of having to manually keep a user heirline config in sync with the core one. I thought it might be best to include it in core given the plugin's popularity (especially considering edgy and Nvim-tree are there as well).

## ℹ Additional Information
**Before**
image
**After**
image


mehalter (approved):
LGTM!

PR #2243, opened 2023-08-27
chore(astro): show "W" before search count in cmdline when wrapping

Subjective:
## 📑 Description

Shows the "W" character before the search count when wrapping around the document. Only applies when cmdheight=1.
## ℹ Additional Information

image


mehalter:
This is a good thing to go in the user configuration

PR #2244, opened 2023-08-27
fix(nvim-ts-autotag): TS closing tag disabled until issue is closed

jay-babu:
## 📑 Description

bug: Autocompletion with slash bug with tsx. Typing something like `<div/`, will result in `<div/div>`, which is invalid JSX. This disables that functionality until it is fixed.

## ℹ Additional Information

https://github.com/windwp/nvim-ts-autotag/issues/125

### DIFF
--- a/lua/plugins/treesitter.lua
+++ b/lua/plugins/treesitter.lua
@@ -3,7 +3,7 @@ return {
   dependencies = {
     "JoosepAlviste/nvim-ts-context-commentstring",
     "nvim-treesitter/nvim-treesitter-textobjects",
-    -- FIXME: remove when https://github.com/windwp/nvim-ts-autotag/issues/125 closed.
+    -- HACK: remove when https://github.com/windwp/nvim-ts-autotag/issues/125 closed.
     {
       "windwp/nvim-ts-autotag",
       opts = {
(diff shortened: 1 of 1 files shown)


lua/plugins/treesitter.lua
```diff
@@ -3,7 +3,16 @@ return {
   dependencies = {
     "JoosepAlviste/nvim-ts-context-commentstring",
     "nvim-treesitter/nvim-treesitter-textobjects",
-    "windwp/nvim-ts-autotag",
+    -- FIXME: remove when https://github.com/windwp/nvim-ts-autotag/issues/125 closed.
```
mehalter, line 6:
```suggestion
    -- HACK: remove when https://github.com/windwp/nvim-ts-autotag/issues/125 closed.
```

I tend to use `FIXME` when there is a known issue/edge case or something incomplete. I think `HACK` is a better annotation here

mehalter (changes requested):
Small comment adjustment. Also please reword the commit message to include the scope. We can use `fix(nvim-ts-autotag):`

mehalter:
Also please fill out the pull request template with the description of the issue and what the fix does, and move the linked issue to the additional information. This is for future reference of people looking back.

mehalter (changes requested):
Thanks for those changes @jay-babu ! Once the pull request description and stuff gets filled out I can get this merged in!

jay-babu:
sorry, was driving, so couldn't do it. should be good now

PR #2251, opened 2023-08-30
chore(ui): set indent_blankline excluded filetype/buftype/bufname on startup

Subjective:
## 📑 Description
 
This sets the excluded filetypes/buftypes/bufnames for indent blankline on startup. The main use case is that they can now be used to configure the exclusions for mini.indentscope (which relies on using an autocmd to disable/enable itself), without sacrificing on lazy-loading.

Also adds the missing buftype exclusions included in the default config for indent_blankline.

## ℹ Additional Information
See https://github.com/AstroNvim/astrocommunity/pull/545


mehalter:
Feel free to do this in your personal configuration if you want. We would prefer to use the `opts` table to align with the configuration of every other plugin.

Subjective:
This was mainly for a draft pr for https://github.com/AstroNvim/astrocommunity/pull/545, but yeah I realize that it certainly isn't ideal to have this kind of inconsistency. I just haven't been able to find another way to make the excluded filetypes and buftypes available on startup to be used in Lazy's `init` function.

mehalter:
You can set it up separately. It doesn't need to use this list. It is more clear if the lists are not shared.

PR #2252, opened 2023-08-30
feat(ui)!: use `mini.indentscope` for highlighting current context

Subjective:
## 📑 Description

**Problem**: The context highlighting built into `indent blankline` is rather buggy and not entirely accurate in many situations (see below screenshots below). 

**Solution**: `mini.indentscope` doesn't suffer the same drawbacks and is vastly more customizable. It might also be worth mentioning that LazyVim uses indent blankline for static indent guides and mini.indentscope for highlighting context.

It also comes with a number of enhancements:
1. Text objects for selecting everything inside and around an indentation guide (similar to https://github.com/arsham/indent-tools.nvim)
2. Navigating through indent scopes via `[i`/`]i` 
3. The option to indent at cursor location (useful for seeing incremental scopes with horizontal cursor movements)
Screen Recording 2023-08-29 at 6 24 17 PM
4. The ability to set the criteria for indent border (for instance, you can place cursor on function header to get scope of its body, which seems to be the default in other editors like vscode)
5. Cool looking animations, though I'm not that big of a fan and keep them disabled by default

Let me know if you think it should be included in AstroNvim by default (or left up to astrocommunity). This is still a WIP as I'm pretty sure the configuration will have to be done in Lazy's `config` function instead of `opts` and a new file will need to be created under "plugins/configs" to allow the user to extend AstroNvim's base config (though I'm not sure if this will change in v4).

## ℹ Additional Information

These are some examples of the inconsistent behavior in `indent blankline` that are fixed by `mini.indentscope`

**Large functions w/ many lines of code**
Screen Recording 2023-08-29 at 6 10 23 PM

**Incorrect context**
image
image

**Ghosting/flickering when changing lines or undoing an edit**
Screen Recording 2023-07-05 at 7 35 52 PM

### DIFF
--- a/lua/astronvim/mappings.lua
+++ b/lua/astronvim/mappings.lua
@@ -437,6 +437,9 @@ if is_available "nvim-cmp" then maps.n["<leader>uc"] = { ui.toggle_cmp, desc = "
 if is_available "nvim-colorizer.lua" then
   maps.n["<leader>uC"] = { "<cmd>ColorizerToggle<cr>", desc = "Toggle color highlight" }
 end
+if is_available "mini.indentscope" then
+  maps.n["<leader>uI"] = { ui.toggle_buffer_indent_guides, desc = "Toggle indent guides (buffer)" }
+end
 maps.n["<leader>ud"] = { ui.toggle_diagnostics, desc = "Toggle diagnostics" }
 maps.n["<leader>ug"] = { ui.toggle_signcolumn, desc = "Toggle signcolumn" }
 maps.n["<leader>ui"] = { ui.set_indent, desc = "Change indent setting" }
--- a/lua/astronvim/utils/ui.lua
+++ b/lua/astronvim/utils/ui.lua
@@ -172,6 +172,20 @@ function M.toggle_signcolumn(silent)
   ui_notify(silent, string.format("signcolumn=%s", vim.wo.signcolumn))
 end
 
+--- Toggle indent guides
+---@param silent? boolean if true then don't sent a notification
+function M.toggle_buffer_indent_guides(silent)
+  vim.g.indent_blankline_enabled = not vim.g.indent_blankline_enabled
+  pcall(vim.cmd.IndentBlanklineRefresh)
+  vim.g.miniindentscope_disable = not vim.g.miniindentscope_disable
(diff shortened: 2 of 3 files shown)


mehalter:
Sure I am open to getting this change into v4 release in a few months

mehalter:
Just make sure you build this off of the v4 branch and not on v3 since this is a breaking change

mehalter:
If you want this in v3 before then, then this is more suited for AstroCommunity

mehalter:
I saw you changed the PR title, I'm not interested in forcing this change into a non-breaking change and with an option in feature. When this is better to be done through AstroCommunity

Subjective:
@mehalter
> I saw you changed the PR title

Oh wait, I was trying to mark this as a WIP, but ended up changing lol. But yeah, this does seem like a change better suited for v4.

mehalter:
I converted the PR to a draft so you don't need to mention it as WIP

mehalter:
@Subjective I've looked into this since you opened this PR and I'm not really sure that this is the direction that we should go with out of the box AstroNvim. This will highly decrease the easy configuration of the indentation plugin, add a lot of complexity with our own managed auto commands, and a lot of other stuff. I think for now staying with indent blankline out of the box is a good solution and for other users they can set up mini on their own or with AstroCommunity.

mehalter:
Just wanted to let you know before you continue putting this together for v4. If you want to still continue and let me look at it when it's done that's also fine. But just telling you where I'm at currently after spending some time looking at the source code and configuration methods.

Subjective:
@mehalter Yeah, I don't really see any other way to disable on filetype other than having an autocmd. I could open a PR in mini to request passing in disabled filetypes/buftypes as an option, but from what I've read it seems like it was the author's intention to have the user configure that stuff.

BTW, I opened a new PR (https://github.com/AstroNvim/AstroNvim/pull/2253) based on v4 with a simpler config – everything is configured within using an opts table instead of a function. I originally took this out of my own user config and I didn't really like the original which-key description for indentscope's text objects, which is why it was originally configured in a function. Let me know what you think, otherwise I guess we can just leave the configuration for astrocommunity.

Subjective:
I think we should continue this on https://github.com/AstroNvim/AstroNvim/pull/2253

PR #2253, opened 2023-08-30
feat(ui)!: use mini.indentscope for highlighting current context

Subjective:
Continuation of https://github.com/AstroNvim/AstroNvim/pull/2252

## 📑 Description

**Problem**: The context highlighting built into `indent blankline` is rather buggy and not entirely accurate in many situations (see below screenshots below). 

**Solution**: `mini.indentscope` doesn't suffer the same drawbacks and is vastly more customizable. It might also be worth mentioning that LazyVim uses indent blankline for static indent guides and mini.indentscope for highlighting context.

It also comes with a number of enhancements:
1. Text objects for selecting everything inside and around an indentation guide (similar to https://github.com/arsham/indent-tools.nvim)
2. Navigating through indent scopes via `[i`/`]i` 
3. The option to indent at cursor location (useful for seeing incremental scopes with horizontal cursor movements)
Screen Recording 2023-08-29 at 6 24 17 PM
4. The ability to set the criteria for indent border (for instance, you can place cursor on function header to get scope of its body, which seems to be the default in other editors like vscode)
5. Cool looking animations, though I'm not that big of a fan and keep them disabled by default

Let me know if you think it should be included in AstroNvim by default (or left up to astrocommunity). This is still a WIP as I'm pretty sure the configuration will have to be done in Lazy's `config` function instead of `opts` and a new file will need to be created under "plugins/configs" to allow the user to extend AstroNvim's base config (though I'm not sure if this will change in v4).

## ℹ Additional Information

These are some examples of the inconsistent behavior in `indent blankline` that are fixed by `mini.indentscope`

**Large functions w/ many lines of code**
Screen Recording 2023-08-29 at 6 10 23 PM

**Incorrect context**
image
image

**Ghosting/flickering when changing lines or undoing an edit**
Screen Recording 2023-07-05 at 7 35 52 PM


Subjective:
Anyways, this is somewhat unrelated, but it seems that some of the incorrect context highlighting I pointed out in the description is actually caused by how AstroNvim has configured the `context_patterns` option. Maybe we should just use the default config instead?

```lua
context_patterns = {
  "class",
  "return",
  "function",
  "method",
  "^if",
  "^while",
  "jsx_element",
  "^for",
  "^object",
  "^table",
  "block",
  "arguments",
  "if_statement",
  "else_clause",
  "jsx_element",
  "jsx_self_closing_element",
  "try_statement",
  "catch_clause",
  "import_statement",
  "operation_type",
}
```

mehalter (approved):
Thanks! I refactored the implementation a bit to fix a few bugs and also aligned the configuration of the two plugins rather than the user needing to configure both simultaneously. I also moved it to use the AstroCore auto commands table while provides an easier route for users to configure auto commands similar to mappings.

mehalter:
@Subjective not sure if you saw that indent blankline v3 is about to be released. We are probably going to be able to drop `mini.indentscope` before the v4 release just heads up :D

Subjective:
@mehalter Looks pretty cool, though I haven't been able to get scope highlighting working that well on lua files. Do you have a config to test out? 

I also wonder if it's possible to get something similar to the horizontal context from indentscope via hooks.

PR #2259, opened 2023-08-31
feat(astro): add mapping to rename sessions

Subjective:
## 📑 Description

Adds mapping to rename sessions. Useful for saving the "Last Session" under a specific name.

## ℹ Additional Information

None


mehalter:
Hm, I'm not a huge fan of this workflow and it's not super clear that this renames the current session. We have save the current session and we have delete a session which I think is a more clear workflow to users than this. Feel free to put this in your user config though!

PR #2261, opened 2023-09-01
fix(plugins): Load nvim-autopairs on User AstroFile instead of Insert…

Uzaaft:
I saw that Autopairs loads in on mini.starter, even if the InsertEnter event doesn't happen. 
This PR aims to solve that.


mehalter:
Thanks! I backported your commit to v3

PR #2274, opened 2023-09-06
[telescope] - live grep + git files

ValentinM27:
## 📑 Description
Add key bind to search thought git files and use live_grep extension for telescope

Add key bind to search thought git files and use live_grep extension for telescope


mehalter:
These are great to go in a user configuration! We do not want to add more extensions to telescope. Thanks for the recommendation though!

ValentinM27:
> These are great to go in a user configuration! We do not want t

Sorry to have bothered you, have a great day ! 👍

PR #2277, opened 2023-09-07
fix(status provider): Check the buffer is valid to avoid crash

Zeioth:
Hey it's been a while. I hope you are doing well.

## What's the purpose of this PR?
It fixes a crash on heirline when using `git mergetool`.

## How does it work?
**Before patch**: Writing the diff causes heirline to crash.
screenshot_2023-09-07_22-57-29_612382661

**After patch**: When writing the diff, it shows the result correctly. On this point we could press `:q` to complete the case of use. All good.
screenshot_2023-09-07_22-57-56_187045117

## How to test

Set your diff tool to neovim with `git config --global mergetool.tool vimdiff `. Then run `git mergetool` on a repo with merge conflicts.

### DIFF
--- a/lua/astronvim/utils/status/provider.lua
+++ b/lua/astronvim/utils/status/provider.lua
@@ -330,7 +330,13 @@ end
 -- @see astronvim.utils.status.utils.stylize
 function M.unique_path(opts)
   opts = extend_tbl({
-    buf_name = function(bufnr) return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":t") end,
+    buf_name = function(bufnr)
+      if vim.api.nvim_buf_is_valid(bufnr) then
+        return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":t")
+      else
+        return ""
+      end
+    end,
     bufnr = 0,
     max_length = 16,
   }, opts)


mehalter:
I actually can't replicate this issue on AstroNvim so I don't think this is necessary. Thanks for the recommendation though!

Zeioth:
Hey no problem! I've been researching for the sake of documenting the issue, so I'm gonna write a bit more about it

## Why it happens
It happens when using the `save session on write` autocmd of [nvim-session-manager](https://github.com/Shatur/neovim-session-manager#save-session-on-save-buffer). This autocmd call the function `session_manager.save_current_session()` which closes any buffer that cannot be saved using `:mksession`. Like for example, a diff :) 

Because the diff buffer is closed as soon as you write it, the provider `unique_path()` is not able to process it fast enough, producing the error `invalid buffer id` (because it doesn't exist anymore).

> The patch fixes this scenario but it is quite a border case

Astro uses a custom solution for session management, so it might not be an issue for users unless they decide to use a different session manager.

## Stack call

```
heirline > winbar > status.component.separated_path() > status.componen.separated_paths() > init.separated_paths > provider.unique_path()
```

mehalter:
Thanks for the detailed writeup! I have tried to replicate this with `nvim-session-manager` which is the default session manager for AstroNvim and cannot replicate the issue. Glad this was able to fix it in NormalNvim, but it looks like AstroNvim might not be affected by the original problem as I'm able to use `nvimdiff` as the `git mergetool` without error.

mehalter:
oh I see @Zeioth it is because you have set up your own auto command to save sessions on writing buffers. I noticed you changed it to saving before leaving neovim. This is actually automatic and you can just delete the auto command entirely. AstroNvim has never used it's own autocommand for that since neovim-session-manager handles that itself. That would explain why you guys have that bug! That way, this change is made unnecessary.

Zeioth:
The fix add error control to `provider.unique_path()` to prevent unexpected scenarios like this, or any other.

I'm glad it's not affecting Astro as the moment.
I keep going with my stuff.
Cheers.

PR #2280, opened 2023-09-09
fix(provider): invalid buffer id in terminal buffer

Subjective:
## 📑 Description

Silences errors when exiting vertical/horizontal split terminal opened manually (e.g. `:horizontal terminal`)

## ℹ Additional Information
Screen Recording 2023-09-09 at 6 36 42 AM

### DIFF
--- a/lua/astronvim/utils/status/provider.lua
+++ b/lua/astronvim/utils/status/provider.lua
@@ -330,7 +330,7 @@ end
 -- @see astronvim.utils.status.utils.stylize
 function M.unique_path(opts)
   opts = extend_tbl({
-    buf_name = function(bufnr) return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":t") end,
+    buf_name = function(bufnr) return vim.fn.fnamemodify(pcall(vim.api.nvim_buf_get_name, bufnr), ":t") end,
     bufnr = 0,
     max_length = 16,
   }, opts)


Subjective:
Perhaps this isn't this best way to solve this, I still get an error message when exiting neovim:
image

mehalter (changes requested):
This shouldn't be done in the `buf_name` function but rather when it is used. `buf_name` is a parameter/argument potentially provided by the user. Could you please refactor this solution or I can do it sometime next week.

mehalter:
This this is not the ideal solution as this doesn't fix the problem you are running into. I would prefer we leave these errors actually so that we don't mask the real issue which would be the buffer management

mehalter:
ohhhhhhh this is a bug only in neovim nightly @Subjective this has to do with a change in behavior that is probably considered a bug in neovim and not actually a bug in AstroNvim

mehalter:
From `:h news`:

```
• Terminal buffers started with no arguments (and use 'shell') close
  automatically if the job exited without error, eliminating the (often
  unwanted) "[Process exited 0]" message.
```

The automatic closing doesn't seem to be triggering our buffer management auto commands. I'll have to see if that's a neovim error or if there is some other auto command event we need to be listening to.

But if you just go to normal mode `<C-\><C-n>` and then run `:q` then it works completely normally

mehalter:
also `:horizontal`/`:vertical` is completely optional in this. You can replicate the same issue with just `:terminal`. Just to help guide this issue and fine tune what's really going on.

Subjective:
Weirdly enough, this only happens for me with `:horizontal` and `:vertical`

mehalter:
I can replicate without

Subjective:
nvm, it also happens without those, just not for a "[No Name]" buffer when opening neovim w/o arguments

mehalter:
@Subjective this is resolved on `nightly`

Subjective:
wow that was quick lol

Subjective:
Might be somewhat related, I get a the same error when closing out of the last window containing a "real" buffer when there is a quickfix window open elsewhere.

Screen Recording 2023-09-09 at 5 57 27 PM

PR #2286, opened 2023-09-13
feat(utils): add `is_loaded` to check whether a plugin has been loaded

Subjective:
## 📑 Description

I wanted to create a heirline component that reports the status of a plugin, but I didn't want it to require the plugin if it wasn't already loaded since that completely messes up lazy-loading. It would be pretty convenient to include a helper function that checks this so that it can be used as a heirline condition.

## ℹ Additional Information

None

### DIFF
--- a/lua/astronvim/utils/init.lua
+++ b/lua/astronvim/utils/init.lua
@@ -223,6 +223,17 @@ function M.is_available(plugin)
   return lazy_config_avail and lazy_config.spec.plugins[plugin] ~= nil
 end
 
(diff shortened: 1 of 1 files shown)


mehalter:
I think it's better to just use `packages.loaded` core table than adding more abstractions, especially more expensive ones like this.

Subjective:
🤦 I was literally just thinking that it might be better to include this in `status.condition` when I saw `function M.aerial_available() return package.loaded["aerial"] end` and realized that was a thing.

PR #2291, opened 2023-09-15
fix(plugins): load astronvim resession extension

Subjective:


mehalter:
Ah this is a mistake. It's `astronvin`

mehalter:
Oh wait I'm an idiot 😂

mehalter:
For records sake, I reverted this on `v4` in favor of renaming the resession extension to `astrocore` to make it more clear where it is coming from.

PR #2300, opened 2023-09-24
feat(mappings): Added keybindings for swapping lines above or below

Bha-Gu:
## 📑 Description
Added keybindings for swapping current line with the line above or below.

## ℹ Additional Information
`[e` swaps the current line with the line above,
`]e` swaps the current line with the line below,

Note: Although, I could not find any of the following but:- The keybinding may be used elsewhere or may be planned for a future use case or; The same functionality may be provided by a different keybinding.


mehalter:
Thanks so much for the recommendation and taking the time to open a PR! I don't think this is the implementation we would want to use here for a couple of reasons. We have sorta stayed away from this binding because the standard binding most people use is `alt-j/k` and I've tried to stay away from binding `alt` in AstroNvim. Also this will not respect indentation like other plugins that implement this will. If you like `]/[e` feel free to add this to your user configuration!

PR #2373, opened 2023-11-19
fix(plugins): update `nvim-ts-context-commentstring` config

salty-flower:
## 📑 Description

A [breaking change](https://github.com/JoosepAlviste/nvim-ts-context-commentstring/commit/5b02387b28a79c61b7d406c2a33d4db1d8454f53) happened and we have to make changes to `JoosepAlviste/nvim-ts-context-commentstring` configurations as guided by the author. See [the README](https://github.com/JoosepAlviste/nvim-ts-context-commentstring#getting-started) for more info.

This commit fixes below warning and associated trace-back notification for me:
> context_commentstring nvim-treesitter module is deprecated, use use require('ts_context_commentstring').setup {} and set vim.g.skip_ts_context_commentstring_module = true to speed up loading instead.
> This feature will be removed in ts_context_commentstring version in the future


mehalter:
We have already done this actually in v4 a couple weeks ago:

https://github.com/AstroNvim/AstroNvim/commit/b0f34d4bca9b4ac330a04ebfc6b3ddf91927ccca

This would constitute a breaking change in v3 only because we have still maintained the ability to use Neovim 0.8. The v4 release will be dropping this backwards compatibility with 0.8.

PR #2376, opened 2023-11-19
fix(options): don't concatenate boolean in error message

wroyca:
## 📑 Description

This fixes what seems to be a typo where we accidentally try to concatenate a boolean with a string in options error message.


mehalter (approved):
LGTM! Thanks!

PR #2381, opened 2023-11-22
fix(plugin): force tabline option 

wroyca:
Heirline is always displayed as an empty winbar area, but tabs are not drawn when 'showtabline' is 0, causing Heirline to show an empty, buggy, and flickering winbar.

The solution seems "obvious" - disable Heirline - but this component is very tightly coupled with Astronvim; disabling it will definitely break something somewhere, and suddenly all hell breaks loose.

So let's do the next "best" thing, which is enforcing 'showtabline,' disregarding user options. If they really want to disable Heirline, they shall do so by overriding the plugin configuration but at this point, they will be held responsible. \;)

### DIFF
--- a/lua/astronvim/options.lua
+++ b/lua/astronvim/options.lua
@@ -28,7 +28,6 @@ vim.opt.relativenumber = true -- show relative numberline
 vim.opt.shiftwidth = 2 -- number of space inserted for indentation
 vim.opt.shortmess:append { s = true, I = true } -- disable search count wrap and startup messages
 vim.opt.showmode = false -- disable showing modes in command line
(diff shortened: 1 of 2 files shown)


mehalter:
This is not what we want. Especially when I'm a prime use case as someone who doesn't use tabline set to 2 😂

wroyca:
> This is not what we want. Especially when I'm a prime use case as someone who doesn't use tabline set to 2 😂

Fair enough, but the bug remains if a user accidentally sets tabline to 0, so should I open a dedicated issue that we can track instead?

mehalter:
No, this isn't a bug. We can discuss on Discord. But I do not get a buggy/flickering winbar

PR #2385, opened 2023-11-25
feat(gitsigns): use histogram algorithm

wroyca:
## 📑 Description

Let's use histogram over Myers' algorithm with gitsigns hunks, as per mentioned on Discord.


wroyca:
Erf. wrong branch.

PR #2386, opened 2023-11-25
feat(gitsigns): use histogram algorithm

wroyca:
## 📑 Description

Let's use histogram over Myers' algorithm with gitsigns hunks, as per mentioned on Discord.


mehalter:
gitsigns by default derives the algorithm from `diffopt` which will unify it with the other diff features of Neovim. We should make this change through that option rather than separating them.

PR #2397, opened 2023-12-09
astronvim_config

hvenkatramana:


mehalter:
I'm guessing this is accidental

hvenkatramana:
Yes its accidental

PR #2404, opened 2023-12-20
Update cmp.lua

8FordPrefect8:
Removing this fixes tab behavior on empty new lines in braces.

## 📑 Description

Started using AstroVim and was confused that in insert mode sometimes hitting tab causes change in visual mode and jump of cursor. Apparently this is caused by this line in cmp config.

### DIFF
--- a/lua/plugins/cmp.lua
+++ b/lua/plugins/cmp.lua
@@ -87,8 +87,6 @@ return {
           ["<Tab>"] = cmp.mapping(function(fallback)
             if cmp.visible() then
               cmp.select_next_item()
-            elseif luasnip.expand_or_jumpable() then
-              luasnip.expand_or_jump()
             elseif has_words_before() then
               cmp.complete()
             else


mehalter:
This is actually intended behavior and is meant to unify snippet jumping and completion selection to enable a "super tab" button. This is actually recommended code from `nvim-cmp`'s wiki (https://github.com/hrsh7th/nvim-cmp/wiki/Example-mappings#luasnip)

Previously I have tried to move away from this and there is always an uproar with the users because most people like this behavior. I personally am not a fan similar to you, but I think the best way is for us to make this change to our user configuration.

Thanks for the suggestion though!

8FordPrefect8:
Okay, hard for me to imagine that anybody really wants this as this drives me crazy.

Is it possible to only overwrite this one mapping from cmp.lua in my user config and let the other things remain tracking from general astrovim? I read the docs, but it is still unclear to me.

mehalter:
haha I agree. The mapping works great as long as you use snippets fully. If you select a snippet from the completion and then you don't actually jump through the snippet edit points then you will get this funky behavior since LuaSnip is horrible and identifying if the user is still actively editing the snippet or not.

Yeah definitely! We actually have a specific documentation page for this: https://docs.astronvim.com/recipes/cmp/#customize-keybindings

8FordPrefect8:
Ah okay, so maybe I misunderstood being a lua noob.

So the line opts = function(_, opts) loads the default and only specified mappings are overriden and final table is returned. I see.

mehalter:
Yup! I would recommend checking out the `lazy.nvim` documentation to understand how the plugin configuration stuff works.

8FordPrefect8:
It worked :).

Maybe I will dig deeper in the future. Right now I am just happy with the experience of astrovim with few things picked from Astrocommunity. The different self constructed frankenstein configs of vim and nvim never achived that level. Thanks for your efforts.

julianschuler:
Hi, just wanted to add here that this behavior is now fixed in #2429, hitting tab in insert mode will now not jump unexpectedly through the file anymore.

PR #2415, opened 2024-01-12
fix(mappings): fixed gdu binary name on macos, it's gdu-go

RayJameson:
Bugs Fixed:
- gdu name on macos is gdu-go


RayJameson:
I didn't cover case when gdu-go is not installed and gdu is installed (part of coreutils, but different program entirely), so it wouldn't trigger warning in this case, also mapping will try to launch `gdu`

RayJameson:
@mehalter how would you like me to work around `health.lua` abstraction? As for now I can't account for both `vim.fn.has "mac"` and `vim.fn.executable` at the same time

mehalter:
@RayJameson Try this commit on AstroNvim nightly: a7c26c6a0d8322ad656270be19be3ee6003a2f8d

mehalter:
Closed in favor of commit linked above

PR #2429, opened 2024-01-20
fix(cmp): only jump through snippet locally

julianschuler:
Bugs Fixed:
- Hitting tab can jump to previously completed snippet from anywhere in the same file

Other:
This PR changes the snippet jump action to only be executed when the cursor is inside the current snippet.

Previously, when e.g. trying to indent a black line using tab an incomplete jumped through previous snippet would result in an unexpected jump to that snippet.

Please let me know if this was a desired behaviour and the selection between local and global snippet jumps should be configurable instead.

### DIFF
--- a/lua/plugins/cmp.lua
+++ b/lua/plugins/cmp.lua
@@ -87,7 +87,7 @@ return {
           ["<Tab>"] = cmp.mapping(function(fallback)
             if cmp.visible() then
               cmp.select_next_item()
-            elseif luasnip.expand_or_jumpable() then
+            elseif luasnip.expand_or_locally_jumpable() then
               luasnip.expand_or_jump()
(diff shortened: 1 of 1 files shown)


mehalter:
This is definitely desired, thanks! Any idea if this is a new feature? I wonder how we have missed this for so long! 😂

mehalter (approved):

julianschuler:
I had a look at the luasnip code base and it seems to have been first introduced in November of 2021.

mehalter:
Dang! Well it's in AstroNvim nightly, should make it into a release in the next couple days :) thanks again! Please feel free to test it by setting your AstroNvim to nightly and make sure it works as intended! I'll do some testing early next week as well before release.

PR #2475, opened 2024-02-21
refactor(utils): create find_config_files function as file helper utility

orhun:
I needed this function in my custom configuration but it was not possible to call it. So I created a new utils module named `file` and moved it to there.

Let me know if you prefer to have it somewhere else.

Thanks!

### DIFF
--- a/lua/astronvim/mappings.lua
+++ b/lua/astronvim/mappings.lua
@@ -255,28 +255,8 @@ if is_available "telescope.nvim" then
   maps.n["<leader>f'"] = { function() require("telescope.builtin").marks() end, desc = "Find marks" }
   maps.n["<leader>f/"] =
     { function() require("telescope.builtin").current_buffer_fuzzy_find() end, desc = "Find words in current buffer" }
-  maps.n["<leader>fa"] = {
-    function()
-      local cwd = vim.fn.stdpath "config" .. "/.."
-      local search_dirs = {}
-      for _, dir in ipairs(astronvim.supported_configs) do -- search all supported config locations
-        if dir == astronvim.install.home then dir = dir .. "/lua/user" end -- don't search the astronvim core files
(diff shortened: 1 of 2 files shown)


mehalter:
Thanks for the recommendation! Sadly, this function will be removed either way on April 1, 2024 (in like 4 weeks) because it is not applicable to the upcoming AstroNvim v4 release. For this reason, I wouldn't recommend upstreaming this at the moment to the utilities nor using any such utility in your user configuration. As a work around, you actually can access that function from within your user config. Here is an example `user/init.lua` file that access it:

```lua
return {
  mappings = function(maps)
    local find_config_files = maps.n["<leader>fa"][1]
    -- do whatever you need...
    
    return maps
  end
}
```

orhun:
Ah I see! Thanks for sharing the solution. Looking forward to the v4 update - I hope the migration won't be painful!

mehalter:
I also hope this 😅 it definitely will be a bit involved haha

orhun:
haha let's see 😄 (💀 💀 💀 )

mehalter:
If you are curious, check out the announcements on the official Discord server. Yesterday we finally got out the first draft of the migration guide :)

orhun:
ok just checked and it indeed looks painful ☠️ I think I'm going to stick with v3 for a while until it breaks. My humble opinion is that it is annoying to re-setup everything while migrating to new versions. However, as a fellow open source developer I see that you're working on improvements that might need that - so I completely understand it. Just wanted to rant here a bit and say that it is just a bit annoying :D

Thanks for your work on this nice neovim config :3

mehalter:
Yeah I really hate it 😅 This is the first really huge breaking change that requires re-setup which is unfortunate but sorta unavoidable in this situation to future proof the project and maintain stability. Our biggest thing is definitely trying to hold off as much as possible any sort of breaking changes to minimize this type of annoyance. Luckily we were able to hold off any breaking changes (v3 -> v4) for over a year at least 😅

PR #2479, opened 2024-03-02
fix(neo-tree): faster open triggered by o

initrc:
## 📑 Description

- Opening a file in neo-tree with `o` is much slower than opening with `l` or `enter`.
- Create an explicit `open` command triggered by `o`. It's as performant as the `child_or_open` command triggered by `l`.


mehalter:
This is not the approach we want to take to solve this. This is resolved in v4 as we are no longer overriding the newly default `o` mappings

PR #2484, opened 2024-03-07
feat(options): enable confirm by default

wroyca:
## 📑 Description

Enable [confirm](https://neovim.io/doc/user/options.html#'confirm') by default so that certain operations that would normally fail because of unsaved changes to a buffer, e.g., ":q" and ":e", instead raise a dialog asking users if they wish to save the current file(s)

## Additional Information

image


mehalter (changes requested):
This is great, could you please move the line to fit into the alphabetically sorted order?

mehalter (approved):
LGTM!

PR #2485, opened 2024-03-07
fix(which-key): handle cmdheight properly

wroyca:
## 📑 Description

When WhichKey draws its trail in cmdline area, having cmdheight set to 0 leads to an incorrect state where both wk's trails and heirline are not properly drawn. The strategy here is to intercept wk.show to toggle cmdheight state. That is, temporarily set cmdheight to 1 during wk.show execution and use a scheduled callback to reset it back to 0 when we're done.

### Before
[Screencast from 2024-03-07 11-48-34.webm](https://github.com/AstroNvim/AstroNvim/assets/91024200/7a39c80f-429e-4910-bfbe-616799ad6849)

### After
[Screencast from 2024-03-07 11-49-55.webm](https://github.com/AstroNvim/AstroNvim/assets/91024200/f0eb8337-c1ca-43a3-83ac-bfc170269460)

### DIFF
--- a/lua/astronvim/plugins/which-key.lua
+++ b/lua/astronvim/plugins/which-key.lua
@@ -5,4 +5,16 @@ return {
     icons = { group = vim.g.icons_enabled ~= false and "" or "+", separator = "" },
     disable = { filetypes = { "TelescopePrompt" } },
   },
+  config = function(_, opts)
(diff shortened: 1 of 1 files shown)


mehalter:
This is incorrect. It shows the trail properly as seen in your recordings

PR #2526, opened 2024-04-06
Close neo-tree if its open before saving session with `resession`

abayomi185:
## 📑 Description

Neovim with `resession` resumes to a blank buffer whenever `neo-tree` is open whilst session was saved (when quitting neovim).
> I seem to remember this issue in V3 at some point but have not seen this occur for the past few months.

- [x] Add pcall to close all `neo-tree` in `resession.lua`

## ℹ Additional Information

Seems like this could be unrelated config `neo-tree` spilling into `resession.lua`. My other thought was with autocmds but I played around with putting an autocmd in `init.lua` presumably to get it loaded first but it didn't seem to work out.

### DIFF
--- a/lua/astronvim/plugins/resession.lua
+++ b/lua/astronvim/plugins/resession.lua
@@ -32,6 +32,8 @@ return {
               local buf_utils = require "astrocore.buffer"
               local autosave = buf_utils.sessions.autosave
               if autosave and buf_utils.is_valid_session() then
+                -- Safely attempt to close Neo-tree if it's open
+                pcall(function() require("neo-tree").close_all() end)
                 local save = require("resession").save
                 if autosave.last then save("Last Session", { notify = false }) end
                 if autosave.cwd then save(vim.fn.getcwd(), { dir = "dirsession", notify = false }) end


mehalter:
This actually doesn't resolve the issue. This is because of the auto quit behavior of `neo-tree`. When you do `<Leader>q` on a buffer with `neo-tree` open it closes the _window_ and not necessarily the editor. So it is closing the window of the file, then it notices that there is only one window left that is `neo-tree`, then it issues a command to quit. That quitting is what triggers the auto saving which is only a `neo-tree` window which is not a valid buffer for saving. So adding this line that you are recommending won't change that the only buffer that is open when the resession saving happens is still just `neo-tree` and it will still result in an empty buffer on restore.

If you have a session open with `neo-tree` open you can actually use `<Leader>Q` with a capital `Q` which issues the command `:qall` which actually will issue a command to exit vim (i.e. quit all windows). This will correctly save the session and it restores without the `neo-tree` window as you would expect.

Sorry for the confusion, I hope this clears up what is happening a bit! Also, sorry that we can't really do anything with the auto-close behavior of neo-tree and how we can hook in a session manager :/ I will make a push to clear up the language of the `<Leader>q` and `<Leader>Q` mappings!

abayomi185:
Ah, thanks for the explanation! It's much clearer to me.

PR #2541, opened 2024-04-16
docs: improve badge layout

uncenter:
## 📑 Description

I believe the whitespace in and around the HTML for the badges as is causes these visual issues where link underlines are visible. This PR compacts the HTML to where that is no longer an issue - and the layout of the badges still appears the same.

Screenshot 2024-04-16 at 10 50 32 (CleanShot)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,30 +17,16 @@
 </h4>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


mehalter (approved):
Nice! Thanks so much!

PR #2564, opened 2024-05-01
doc: update the update command

ALameLlama:
## 📑 Description

it looks like with the release of 4.0, AstroUpdatePackages was removed and AstroUpdate does the same thing (it calls astrocore and updates everything)


mehalter (approved):
Thanks!

PR #2573, opened 2024-05-04
refactor(options): set shiftwidth to automatically copy 'tabstop' value

wroyca:
## 📑 Description

When ['shiftwidth'](https://neovim.io/doc/user/options.html#'shiftwidth') is zero, the ['tabstop'](https://neovim.io/doc/user/options.html#'tabstop') value will be used, so we can avoid one extra step when changing the number of spaces that a `<Tab>` in the file counts for


mehalter (approved):
Looks great! Thanks!

PR #2599, opened 2024-05-15
chore(colorizer): replaced nvim-colorizer with nvim-highligt-colors

Std-Enigma:


mehalter:
This is done on a branch here already that is though!

mehalter:
For reference, here is the branch with `nvim-highlight-colors`: https://github.com/AstroNvim/AstroNvim/tree/highlight_colors

PR #2623, opened 2024-06-06
fix(main): update description of refactoring menu

Infonautica:
## 📑 Description

- Removed a duplicate mapping for "gra"
- Added human readable descriptions for the items of refactoring menu that were added recently

## ℹ Additional Information

I'm talking about these:
Screenshot 2024-06-07 at 00 16 22

### DIFF
--- a/lua/astronvim/plugins/_astrocore_mappings.lua
+++ b/lua/astronvim/plugins/_astrocore_mappings.lua
@@ -56,10 +56,9 @@ return {
 
     -- Neovim Default LSP Mappings
     if vim.fn.has "nvim-0.11" ~= 1 then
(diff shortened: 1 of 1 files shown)


mehalter (commented):
These are in place actually to match the core neovim mappings that are being added so I think we might want to leave them as is

Infonautica:
> we might want to leave them as is

That's alright, what about double definition of "gra"?

mehalter:
That isn't a duplicate. One is for normal mode and the other is for visual mode

Infonautica:
> That isn't a duplicate. One is for normal mode and the other is for visual mode

Thanks for catching that 👏

PR #2637, opened 2024-06-24
feat(telescope): open multiple selected files

kedlubnowski:
## 📑 Description
This PR allows to hit `CR` key to open multiple files from telescope selected via `Tab` key and jump to the selected row if applicable. If the file is selected more then once, it opens only once and jumps to the row from the last selected entry. Same if the file is already opened.
It also adds the `M-CR` (Alt-Enter) key to open all files showed in the telescope.

I took  [this code](https://github.com/nvim-telescope/telescope.nvim/issues/1048#issuecomment-2142669167) as a base and slightly modify it.


kedlubnowski:
Are you shure that C-CR will work? Because that was my first attempt and it didn't work for me. It act as a simple CR.

mehalter:
Yeah it still works, although it might take a modern terminal to support that binding. You can test if your terminal supports it by going into insert mode and pressing `<C-V><C-CR>` and it should type out `<C-CR>`. Your terminal also could be consuming that binding before ever getting to neovim if it's mapped to something else externally

mehalter:
ah yeah, looking now it looks like it's a binding most terminal emulators just don't support. I'm down to change it to something else not alt related or to just remove that binding. I think the amount of times people want to open everything in telescope is probably slim to none and if they do they can select all of them. Maybe it's better to just make a bidning to the `select_all` action that would empower opening everything as well as sending everything to the quickfix

kedlubnowski:
I prefer to change it to something non-conflicting rather than removing it.

kedlubnowski:
Ok, thanks!

PR #2710, opened 2024-09-24
fix(telescope): ensure `.git` folders are always ignored on Windows

AnsonH:
## 📑 Description

A follow-up of 016c2cfc5288118f315a7fb26082d90d86a529b6 by ensuring `.git` folders are properly excluded from Telescope search results on Windows machines.

The original Regex file patterns of `"^%.git/"` and `"/%.git/"` could not match Git folders on Windows because Windows uses backward slashes (`\`) instead of forward slashes. 

## ℹ Additional Information

Steps to reproduce:
1. Install AstroNvim on a Windows machine
2. Clone this repo
3. Run `:Telescope find_files`
4. Search for the word `.git`

- Before: `.git\*` files will show up
- After: The Git files are hidden

| Before | After |
| ------ | ----- |
| image | image

### DIFF
--- a/lua/astronvim/plugins/telescope.lua
+++ b/lua/astronvim/plugins/telescope.lua
@@ -124,7 +124,7 @@ return {
     end
     return {
       defaults = {
-        file_ignore_patterns = { "^%.git/", "/%.git/" },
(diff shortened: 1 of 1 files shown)


mehalter (approved):
LGTM! thanks!

PR #2713, opened 2024-09-27
fix(mapping): shorten the description of gx

zhou13:
## 📑 Description

The description of `gx` is too long and `which-key` can only display in 1-col mode. This PR fixes this.

Before:
Screenshot 2024-09-27 at 2 25 46 PM

After
Screenshot 2024-09-27 at 2 26 11 PM

### DIFF
--- a/lua/astronvim/plugins/_astrocore_mappings.lua
+++ b/lua/astronvim/plugins/_astrocore_mappings.lua
@@ -42,7 +42,7 @@ return {
     maps.n["\\"] = { "<Cmd>split<CR>", desc = "Horizontal Split" }
     -- TODO: remove deprecated method check after dropping support for neovim v0.9
     if not vim.ui.open then
-      local gx_desc = "Opens filepath or URI under cursor with the system handler (file explorer, web browser, …)"
+      local gx_desc = "System-open path under cursor"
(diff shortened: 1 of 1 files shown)


mehalter (commented):
This will not fix this. We only create this mapping in Neovim v0.9. In Neovim v0.10 this is a core, default mapping. We are simply using the actual default mapping. If you are using Neovim v0.10 then this wouldn't change the description since we don't set the mapping

mehalter:
We are just going to stick with this since we want to stick as close to stock neovim as possible and only change what's really necessary

zhou13:
I am on `NVIM v0.10.1` and I am still getting this.

mehalter:
Yeah, so even if I merged this in you would still get the long description in Neovim v0.10.1 because that's what Neovim puts it at and we do not create this mapping. This PR would _only_ affect Neovim v0.9

mehalter:
https://github.com/neovim/neovim/blob/master/runtime/lua/vim/_defaults.lua#L113-L132

Here is the core neovim code that sets the mapping with the description. The only reason we have it in the astronvim source code at the moment is to backport this directly to neovim v0.9. In the next major release of AstroNvim this mapping is going to be removed from our source code all together and we will only rely on neovim to supply it

mehalter:
You could totally add something to your own configuration to shorten the description! We just don't want to merge it to core AstroNvim

PR #2722, opened 2024-10-27
feat(gitsigns): close preview_hunk window with q

windowsrefund:
## 📑 Description

Closing the `preview_hunk` window is normally done by moving the cursor. I've found this to be disruptive at times so went ahead and added a mapping to close the window with `q`.  Of course, one could argue this is the wrong place to add such a binding as closing any window like this via a standard key like `q` would offer the most benefit being scoped elsewhere. Obviously, that's a call for the Jedi to make...

Scooped this little gem [here](https://github.com/lewis6991/gitsigns.nvim/issues/385#issuecomment-1635543223)


windowsrefund:
Actually, it would be nice to find a way to scope this higher as it'd be nice to have this available as a means to exit from a hover. I really wouldn't know how to attack that though.

PR #2732, opened 2024-11-22
Update default values for diagnostics.

gwww:
Use default for `update_in_insert`.

Fix `source` to a value defined by type.


mehalter:
Merged manually in v5: https://github.com/AstroNvim/AstroNvim/commit/b7d1865281f32c26c63900934cccf89090aaf1a6

PR #2758, opened 2025-02-19
fix(_astrolsp): use correct key when indexing file_operations table

LilDojd:
## 📑 Description

Replace the incorrect reference of `module` with `operation` in the file operations event handler.

- [x] Fixed variable reference from `module` to `operation` in the handler function.

## ℹ Additional Information
--


mehalter (approved):
LGTM

PR #2792, opened 2025-04-15
feat(snacks): add undo history picker

ALameLlama:
## 📑 Description

Added the snacks undo history picker, which uses the same keybind as the teloscope plugin 
https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/editing-support/telescope-undo-nvim/init.lua


mehalter (approved):
Looks great!

PR #2794, opened 2025-04-15
chore(main): remove telescope referances

ALameLlama:
## 📑 Description

Noticed it still says telescope in some places when doing a health check so I updated them


ALameLlama:
I also when when he'll add a command for TodoPicker or something like he did with TodoTelescope and TodoFzf
https://github.com/AstroNvim/AstroNvim/blob/main/lua/astronvim/plugins/todo-comments.lua#L4

PR #2798, opened 2025-04-16
chore: Update macOS version in issue template

LandonSchropp:
## 📑 Description

This updates the version of macOS in the issue template. This probably isn't worth a PR, but YOLO.


mehalter:
Idk this doesn't change the functionality of the template or anything. It doesn't even change the default value as the user would still have to type version 15 if they were using it 😅 so I don't think we need to make this change

PR #2799, opened 2025-04-17
fix(highlight-colors): fix typo in config option

ALVAROPING1:
## 📑 Description

The `highlight-colors`'s config was trying to disable `enabled_named_colors`, but according to the plugin's docs it's spelled as `enable_named_colors` (without a final `d` in `enable`) which lead to the config being ignored. This PR fixes the name of the setting


mehalter (approved):
Looks great! Thanks!

PR #2817, opened 2025-05-27
feat(neo-tree): integration Snacks.rename

msdone-lwt:
## 📑 Description
Integrate Neo-Tree with the Snacks.rename feature.


mehalter:
We do not want to use snacks rename. AstroNvim has their own implementation of this that is more correct. Thanks though!

mehalter:
If you are curious here is the LSP file action features in AstroLSP: https://github.com/AstroNvim/astrolsp?tab=readme-ov-file#lsp-file-operations

PR #2842, opened 2025-08-26
feat(heirline): make close button optional via astroui status config

danielfalbo:
This PR adds the ability to disable the close button in the tabline via the astroui status configuration.

The close button can be disabled by setting `show_close_button = false` in the astroui status configuration. By default, it is enabled to maintain backward compatibility.

Changes:
- Added `show_close_button` option to astroui status configuration (default: true)
- Modified heirline tabline to respect the configuration option

See https://github.com/AstroNvim/astroui/pull/63

### DIFF
--- a/lua/astronvim/plugins/_astroui.lua
+++ b/lua/astronvim/plugins/_astroui.lua
@@ -36,6 +36,9 @@ return {
   ---@type AstroUIOpts
   opts = {
     colorscheme = "astrotheme",
+    status = {
+      show_close_button = true, -- enable close button in tabline by default
+    },
     folding = {
       enabled = function(bufnr) return require("astrocore.buffer").is_valid(bufnr) end,
       methods = { "lsp", "treesitter", "indent" },
--- a/lua/astronvim/plugins/heirline.lua
+++ b/lua/astronvim/plugins/heirline.lua
@@ -128,14 +128,16 @@ return {
             provider = status.provider.tabnr(),
             hl = function(self) return status.hl.get_attributes(status.heirline.tab_type(self, "tab"), true) end,
           },
-          vim.tbl_get(ui_config, "status", "show_close_button") ~= false and { -- close button for current tab
-            provider = status.provider.close_button { kind = "TabClose", padding = { left = 1, right = 1 } },
-            hl = cached_func(status.hl.get_attributes, "tab_close", true),
-            on_click = {
-              callback = function() require("astrocore.buffer").close_tab() end,
-              name = "heirline_tabline_close_tab_callback",
-            },
-          } or nil,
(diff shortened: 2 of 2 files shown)


mehalter:
Thanks for the recommendation! In terms of this type of configuration for building statusline/tabline/winbar. We are opting to stay far away from big abstractions like this and for users who want to customize their full bars/lines they need to manage their own code for this. For example, you can disable the tabline tab close button in a user configuration with:


```lua
return {
  "rebelot/heirline.nvim",
  opts = function(_, opts)
    local status = require "astroui.status"
    opts.tabline = { -- bufferline
      { -- automatic sidebar padding
        condition = function(self)
          self.winid = vim.api.nvim_tabpage_list_wins(0)[1]
          self.winwidth = vim.api.nvim_win_get_width(self.winid)
          return self.winwidth ~= vim.o.columns -- only apply to sidebars
            and not require("astrocore.buffer").is_valid(vim.api.nvim_win_get_buf(self.winid)) -- if buffer is not in tabline
        end,
        provider = function(self) return (" "):rep(self.winwidth + 1) end,
        hl = { bg = "tabline_bg" },
      },
      status.heirline.make_buflist(status.component.tabline_file_info()), -- component for each buffer tab
      status.component.fill { hl = { bg = "tabline_bg" } }, -- fill the rest of the tabline with background color
      { -- tab list
        condition = function() return #vim.api.nvim_list_tabpages() >= 2 end, -- only show tabs if there are more than one
        status.heirline.make_tablist { -- component for each tab
          provider = status.provider.tabnr(),
          hl = function(self) return status.hl.get_attributes(status.heirline.tab_type(self, "tab"), true) end,
        },
      },
    }
  end,
}
```

If this is too verbose, and sometimes it is a lot of users opt to move to a different statusline provider that meets their customization needs such as lualine which is easily migratable with AstroCommunity as a starting point.

danielfalbo:
This is much more elegant, thank you!
I ended up doing
```lua
return {
  "rebelot/heirline.nvim",
  opts = function(_, opts)
    local status = require "astroui.status"
    opts.tabline[2] = status.heirline.make_buflist(status.component.tabline_file_info({ close_button = false }))
  end,
}

```

danielfalbo:
Ah got it, I didn't know about that because I don't use tabs that much. But yes now I see and I want to remove the mouse-clickable x-mark for the tabs as well `:)`

I updated my config to

```lua
return {
  "rebelot/heirline.nvim",
  opts = function(_, opts)
    local status = require "astroui.status"
    opts.tabline[2] = status.heirline.make_buflist(status.component.tabline_file_info { close_button = false })
    local tab_section = opts.tabline[#opts.tabline]
    tab_section[#tab_section] = nil
  end,
}
```

Thanks again!

PR #2848, opened 2025-09-26
fix(blink): fix missing snippet icon when use `lspkind` symbols

nicholasxjy:
## 📑 Description

Fix no snippet icon in blink.cmp when use `lspkind` symbols

## 📖 Additional Information

- before

Screenshot 2025-09-26 at 11 05 41

- after

Screenshot 2025-09-26 at 11 32 20


mehalter (approved):
Thanks!

PR #2602, opened 2024-05-18
docs: fix incorrect admonition syntax

loichyan:
## 📑 Description

Fixes the incorrect admonition syntax introduced in AstroNvim/AstroNvim@0f875f4a7f1963d6c61eb564cadb8e556e406dd5 :

Snapshot

PR #2887, opened 2026-07-22
docs: fix typos, admonition rendering, and outdated version in README

takano32:
Three small README fixes:

- `[!NOTE]` admonitions were written on a single line together with the following text, so GitHub renders them as literal `[!NOTE] >` text instead of callouts. This restores the fix from #2602, which had regressed.
- Typos: `Some user's` -> `Some users`, `set's up` -> `sets up`, a stray comma inside the logo `` tag, and a misplaced conjunction in the Heirline feature list (`Statusline, Winbar, and Bufferline, Statuscolumn`).
- The minimal `init.lua` example still pins `version = "^5"` while the current major version is v6, so users copying the snippet install the previous major version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ if not vim.uv.fs_stat(lazypath) then
 end
 vim.opt.rtp:prepend(lazypath)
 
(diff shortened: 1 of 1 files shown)


mehalter (approved):
Thanks!

PR #2889, opened 2026-07-22
fix(astrocore): correct Makefile capitalization in rooter detector

takano32:
The rooter fallback detector lists `"MakeFile"`, which never matches a real `Makefile` on case-sensitive filesystems, so projects rooted only by a Makefile are not detected on Linux.


mehalter (approved):
Thanks!

PR #2890, opened 2026-07-22
perf: migrate more vim functions to Lua APIs

takano32:
Continues 2946f98 by migrating the remaining straightforward cases:

- health.lua: `vim.fn.matchstr(vim.fn.execute "version", ...)` -> `vim.api.nvim_exec2("version", { output = true }).output:match(...)`. Output is byte-identical; `tostring(vim.version())` is not a substitute because it appends build metadata on release builds.
- _astrolsp.lua: `vim.fn.expand("#" .. args.buf .. ":p")` -> `vim.api.nvim_buf_get_name(args.buf)` (both return the full path, and `""` for unnamed buffers); matches the idiom already used in _astrocore_autocmds.lua and neo-tree.lua.
- snacks.lua: drop the `(vim.uv or vim.loop)` fallback — `vim.uv` always exists on the minimum supported Neovim 0.11, and this was the last remaining `vim.loop` reference.

### DIFF
--- a/lua/astronvim/health.lua
+++ b/lua/astronvim/health.lua
@@ -13,7 +13,7 @@ function M.check()
   health.start "Checking requirements"
 
   health.info("AstroNvim Version: " .. require("astronvim").version())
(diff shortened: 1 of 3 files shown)


mehalter (approved):
Thanks!

